PR #335, opened 2014-12-11
apirequest/isReadableStream rework

### LINKED ISSUE #334: http.ServerResponse is not recognized as ReadableStream
using GCS, when creating a new object, i tried to directly pipe in a http response, but API does not recognize it as readable stream.

	http.get('http://www.example.com/image.jpeg', function(response){
		storage.objects.insert({bucket: 'my-bucket',name:'testThing',media:{body:response}}, function (err,uploadResult){
			if (err){
				console.log('err',err);
				return res.status(500).send('upload failed');
			}
			return res.status(201).send({id:123456});
		})
	})

Providing a pull request that fixes it for me.

viktorasm:
I tried using a http.get response object directly as GCS media.body, which should be a valid readable stream.

Fixes #334

### DIFF
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -19,7 +19,6 @@
 var Multipart = require('multipart-stream');
 var utils = require('./utils.js');
 var DefaultTransporter = require('./transporters.js');
-var stream = require('stream');
 var parseString = require('string-template');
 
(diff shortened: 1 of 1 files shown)


ryanseys:
Closing as I don't see a need for this.

PR #343, opened 2015-01-06
Use the new Travis infrastructure

Alaneor:
More info:
http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure

Travis now uses Docker-based infrastructure which is more efficient and much faster. Let's upgrade (since we do not need any of the features currently limited to the old VM infra, like `sudo`)!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: node_js
 node_js:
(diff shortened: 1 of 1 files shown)


ryanseys:
Thanks!

PR #346, opened 2015-01-12
All API call errors should be returned as instances of Error

### LINKED ISSUE #345: Callback error should be an Error instance, not a plain object
That is the Node convention and the expected output.

See: https://github.com/addyosmani/psi/issues/43

### Test-case

```js
var googleapis = require('googleapis');
var pagespeed = googleapis.pagespeedonline('v1').pagespeedapi.runpagespeed;

// intentional grammar mistake to trigger the error
pagespeed({url: 'http://google.con'}, function (err, response) {
  if (err) {
    throw err;
    //=> [object Object]
  }
});
```

It should throw the error with stack, but instead just throws `[object Object]`.

Alaneor:
Hopefully I have not missed any other place where we wrap callbacks or create errors ourselves...

Thanks to @sindresorhus for pointing the issue out!
Closes #345

### DIFF
--- a/test/test.transporters.js
+++ b/test/test.transporters.js
@@ -71,6 +71,22 @@ describe('Transporters', function() {
     assert.equal(req.body, null);
   });
 
+it('should return errors within response body as instances of Error', function(done) {
+    var google = require('../lib/googleapis');
+    var drive = google.drive('v2');
+
+    var scope = nock('https://www.googleapis.com')
+      .get('/drive/v2/files?q=hello')
+      // Simulate an error returned via response body from Google's API endpoint
+      .reply(400, { error: { code: 400, message: 'Error!'} });
+
+    drive.files.list({ q: 'hello' }, function(err) {
+      assert(err instanceof Error);
+      scope.done();
+      done();
+    });
+  });
+
   it('should return 5xx responses as errors', function(done) {
     var google = require('../lib/googleapis');
     var scope = nock('https://www.googleapis.com')
@@ -79,7 +95,9 @@ describe('Transporters', function() {
     var urlshortener = google.urlshortener('v1');
     var obj = { longUrl: 'http://google.com/' };
     urlshortener.url.insert({ resource: obj }, function(err, result) {
-      assert.deepEqual(err, { code: 500, message: 'There was an error!' });
+      assert(err instanceof Error);
+      assert.equal(err.code, 500);
+      assert.equal(err.message, 'There was an error!');
       assert.equal(result, null);
       scope.done();
       done();
--- a/lib/transporters.js
+++ b/lib/transporters.js
@@ -75,14 +75,16 @@ DefaultTransporter.prototype.wrapCallback_ = function(opt_callback) {
 
     if (body && body.error) {
       // handle single request errors
-      err = body.error;
+      err = new Error(body.error.message);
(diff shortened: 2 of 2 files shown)


test/test.transporters.js
```diff
@@ -79,7 +90,9 @@ describe('Transporters', function() {
     var urlshortener = google.urlshortener('v1');
     var obj = { longUrl: 'http://google.com/' };
     urlshortener.url.insert({ resource: obj }, function(err, result) {
-      assert.deepEqual(err, { code: 500, message: 'There was an error!' });
+      assert(err instanceof Error);
+      assert.equal(err.code, 500);
+      assert.equal(err.message, 'There was an error!');
```
ryanseys: Can we also test for .stack object here for the stack trace.

Alaneor: Is it necessary...? We are testing if the object is instance of `Error`, that should implicitly test for presence of `stack`. What do you think...?

ryanseys: That's fair. I think as of 0.10.x they put the stack object in the errors by default instead of having to capture it with captureStackTrace. Considering we only support 0.10.x this is a fair assumption.

test/test.transporters.js
```diff
@@ -71,6 +71,17 @@ describe('Transporters', function() {
     assert.equal(req.body, null);
   });
 
+it('should return errors as instances of Error', function (done) {
+    var google = require('../lib/googleapis');
+    var pagespeed = google.pagespeedonline('v1');
+
+    // intentional grammar mistake to trigger the error
+    pagespeed.pagespeedapi.runpagespeed({url: 'http://gogle.con'}, function (err) {
```
ryanseys: Is this making a real request? None of the unit tests should depend on network access.

Alaneor: Yup, it is - I will *nock* it...

ryanseys: :dancer:

ryanseys:
This should require a version bump to 1.1.0

test/test.transporters.js
Alaneor: Actually, how would you simulate a DNS resolution failure? *Nock* sure can't do it. Also, recently I gave [this answer](http://stackoverflow.com/questions/27708960/how-to-test-error-in-request-with-nock/27765293#27765293) over on SO to exactly this kind of problem and I am not sure if I want to go into such monkey-patching in real world...:)

So, maybe to clarify - this test does not try to prevent the network call from happening, but in reality it will never happen because it cannot resolve *http://gogle.con* into IP address.

ryanseys: But we just send the url to google and its servers tries to run the IP resolve, then they return the error when it fails. So we don't have any issue making the request on our end, but google's servers do.

Alaneor: Oooooh :bulb: Of course, of course...!

Alaneor:
Behold, the tests have been updated!

ryanseys:
Sweet, looks good. 1.1.0?

Alaneor:
Probably - we changed how the response body is delivered to API consumers in case of error, so just to be on the safer side.

ryanseys:
Released 1.1.0 https://github.com/google/google-api-nodejs-client/releases/tag/v1.1.0

PR #349, opened 2015-01-14
Do not output stuff to console during testing

Alaneor:
It looks ugly.:)


ryanseys:
Thanks :) Not sure why I didn't think of that earlier! :fearful:

PR #361, opened 2015-02-11
Handle 500 errors that include an error message.

### LINKED ISSUE #360: Not all errors passed through to error message
I'm seeing a lot of errors passed back to my callbacks that include a null message. Eg:

```
{ [Error: null] code: 500 }
```

However, when examining the body of the actual HTTP response, I see that the message is actually populated:

Actual response object:
```
 {
   "statusCode": 500,
   "body": {
     "error": {
       "errors": [
         {
           "domain": "global",
           "reason": "backendError",
           "message": "Backend Error"
         }
       ],
       "code": 500,
       "message": "Backend Error"
     }
   }
 }
```

bradvogel:
Fixes #360

### DIFF
--- a/lib/transporters.js
+++ b/lib/transporters.js
@@ -73,26 +73,25 @@ DefaultTransporter.prototype.wrapCallback_ = function(opt_callback) {
       body = JSON.parse(body);
     } catch (err) { /* no op */ }
 
-    if (body && body.error) {
-      // handle single request errors
+    // Consider the response an error if it includes the 'error' property or it has a
+    // 500 status code.
+    if ((body && body.error) || res.statusCode >= 500) {
 
-      if (typeof body.error === 'string') {
+      if (!body.error){
+        err = new Error(body);
+        err.code = res.statusCode;
+        err.type = body.error;
+
+      } else if (typeof body.error === 'string') {
         err = new Error(body.error_description);
         err.code = res.statusCode;
         err.type = body.error;
+
       } else {
         err = new Error(body.error.message);
-        err.code = body.error.code;
+        err.code = body.error.code || res.statusCode;
       }
 
-      body = null;
-    }
-
-    // there is no err and is a body and no body.error
-    // but server still returns error code
-    if (res.statusCode >= 500) {
-      err = new Error(body);
-      err.code = res.statusCode;
       body = null;
     }
 
--- a/lib/transporters.js
+++ b/lib/transporters.js
@@ -73,15 +73,8 @@ DefaultTransporter.prototype.wrapCallback_ = function(opt_callback) {
       body = JSON.parse(body);
     } catch (err) { /* no op */ }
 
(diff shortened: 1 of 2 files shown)


test/test.transporters.js
```diff
@@ -124,4 +124,22 @@ it('should return errors within response body as instances of Error', function(d
       done();
     });
   });
+
+  it('should handle 5xx responses that include errors', function(done) {
+    var google = require('../lib/googleapis');
+    var scope = nock('https://www.googleapis.com')
+      .post('/urlshortener/v1/url')
+      .reply(500, { error: {message: 'There was an error!'}});
```
ryanseys: Should also test with your original object you were having issues with:

``` js
 {
   "statusCode": 500,
   "body": {
     "error": {
       "errors": [
         {
           "domain": "global",
           "reason": "backendError",
           "message": "Backend Error"
         }
       ],
       "code": 500,
       "message": "Backend Error"
     }
   }
 }
```

lib/transporters.js
```diff
@@ -73,29 +73,28 @@ DefaultTransporter.prototype.wrapCallback_ = function(opt_callback) {
...
-      // handle single request errors
+    // Consider the response an error if it includes the 'error' property or it has a
+    // 500 status code.
+    if ((body && body.error) || res.statusCode >= 500) {
 
-      if (typeof body.error === 'string') {
+      if (!body.error){
+        err = new Error(body);
+        err.code = res.statusCode;
+        err.type = body.error;
```
ryanseys: This is null or undefined.

lib/transporters.js
```diff
@@ -73,29 +73,28 @@ DefaultTransporter.prototype.wrapCallback_ = function(opt_callback) {
       body = JSON.parse(body);
     } catch (err) { /* no op */ }
 
-    if (body && body.error) {
-      // handle single request errors
+    // Consider the response an error if it includes the 'error' property or it has a
+    // 500 status code.
+    if ((body && body.error) || res.statusCode >= 500) {
```
ryanseys: I don't really like this nested if block. I believe we could achieve the same thing with an if-else block

test/test.transporters.js
bradvogel: Added below.

ryanseys: Isn't the whole body the entire object I put above? Not just the body component of that object? I might be mistaken.

bradvogel:
Updated.

test/test.transporters.js
bradvogel: The JSON I pasted (which you have above) is the complete JSON representation of the response object. So only the 'body' part is the JSON returned in the response body.

ryanseys: Ah my bad.

bradvogel:
Thanks! When are you planning to push another version?

ryanseys:
Released.

PR #363, opened 2015-02-13
Pass req_opts from constructor args to request in oauth2client

abrgr:
This permits client code to pass options (e.g. proxy) directly to request (the transport library).

### DIFF
--- a/lib/auth/oauth2client.js
+++ b/lib/auth/oauth2client.js
@@ -21,6 +21,7 @@ var AuthClient = require('./authclient.js');
 var util = require('util');
 var PemVerifier = require('./../pemverifier.js');
 var LoginTicket = require('./loginticket.js');
+var utils = require('../utils');
 
 var certificateCache = null;
 var certificateExpiry = null;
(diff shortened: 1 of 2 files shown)


ryanseys:
Could you write some docs on how I would use this?

ryanseys:
Also, perhaps some simple tests to ensure the parameters are being passed correctly, and makes documenting much easier.

abrgr:
I'll do my best to add the tests later today.

My usage looks like this:

```
var oauthOpts = {
    req_opts: {
        proxy: 'http:/myproxy:80'
    }
};

var oauth2Client = new OAuth2(
    cfg.google.clientId,
    cfg.google.clientSecret,
    cfg.google.clientUrl,
    oauthOpts
);

//...
```

Do you want me to add this to any of the markdown docs or just leave it here and in the tests?

ryanseys:
Oh hmm. Yeah that's a little messy. Can we instead just add support for a `proxy` value in the `oauthOpts` object, rather than munging all the opts together?

abrgr:
I liked the idea of being able to set arbitrary options on request because there are a few that I want set on all external API calls.  In my case, I'm setting a proxy and timeout.  The idea of munging all the opts seemed more in-line with https://github.com/google/google-api-nodejs-client/blob/master/lib/apirequest.js#L179, which allows arbitrary access to the request options.

That said, if you'd rather cherry-pick properties, I'm happy to make that change.  Mind if I add proxy and timeout?

ryanseys:
Yeah, arbitrary request options are available on the API request level but I didn't think about at the transporter level itself. :( In either case `req_opts` should probably be better named and use camel case to fit with our conventions. Perhaps `requestOpts` is a more explicit name to consider.

abrgr:
You got it.  I'll update it to requestOpts and add some tests.

JustinBeckwith:

PR #371, opened 2015-02-19
Batch Requests

ahageali:
While experimental, I have tested this through the use of urlshortener apis and drive apis and this does appear to work for normal GET and POST requests. I have also added some documentation to provide a simple way to call the batch api. As a result however, the api requests no longer actually make their requests when they are called without a callback. 

I have also _tried_ to keep as similar a style as possible to all the pre-existing code so I apologize if I made any horrendous mistakes. Let me know if you want any changes!

### DIFF
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -85,10 +85,11 @@ function createAPIRequest(parameters, callback) {
   var req, body, missingParams;
   var params = parameters.params;
   var options = utils.extend({}, parameters.options);
-
+  var missingCallback;
   // If the params are not present, and callback was passed instead,
   // use params as the callback and create empty params.
   if (typeof params === 'function') {
+    missingCallback = false;
     callback = params;
     params = {};
   }
@@ -103,14 +104,15 @@ function createAPIRequest(parameters, callback) {
   );
 
   // Normalize callback
-  // callback = createCallback(callback);
+  missingCallback = !callback;
+  callback = createCallback(callback);
 
   // Check for missing required parameters in the API request
   missingParams = getMissingParams(params, parameters.requiredParams);
   if (missingParams) {
     // Some params are missing - stop further operations and inform the developer which required
     // params are not included in the request
-    callback && callback(new Error('Missing required parameters: ' + missingParams.join(', ')));
+    callback(new Error('Missing required parameters: ' + missingParams.join(', ')));
 
     return null;
   }
@@ -184,7 +186,7 @@ function createAPIRequest(parameters, callback) {
   delete options.auth; // is overridden by our auth code
   delete options.params; // We handle params ourselves and Request does not recognise 'params'
 
-  if (!callback) {
+  if (missingCallback) {
     return options;
   }
 
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -89,7 +89,6 @@ function createAPIRequest(parameters, callback) {
   // If the params are not present, and callback was passed instead,
   // use params as the callback and create empty params.
   if (typeof params === 'function') {
-    missingCallback = false;
     callback = params;
     params = {};
   }
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -175,6 +175,9 @@ function createAPIRequest(parameters, callback) {
   delete options.params; // We handle params ourselves and Request does not recognise 'params'
 
   if (!callback) {
+    if (body) {
+      options.stream_body = body;
+    }
     return options;
   }
 
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -103,14 +103,14 @@ function createAPIRequest(parameters, callback) {
   );
 
   // Normalize callback
-  callback = createCallback(callback);
+  // callback = createCallback(callback);
 
   // Check for missing required parameters in the API request
   missingParams = getMissingParams(params, parameters.requiredParams);
   if (missingParams) {
     // Some params are missing - stop further operations and inform the developer which required
     // params are not included in the request
-    callback(new Error('Missing required parameters: ' + missingParams.join(', ')));
+    callback && callback(new Error('Missing required parameters: ' + missingParams.join(', ')));
 
     return null;
   }
@@ -183,6 +183,10 @@ function createAPIRequest(parameters, callback) {
   );
   delete options.auth; // is overridden by our auth code
   delete options.params; // We handle params ourselves and Request does not recognise 'params'
+
+  if (!callback) {
+    return options;
+  }
 
   // create request (using authClient or otherwise and return request obj)
   if (authClient) {
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -27,16 +27,6 @@ function isReadableStream(obj) {
     typeof obj._readableState === 'object';
 }
 
-function logError(err) {
-  if (err) {
-    console.error(err);
-  }
-}
-
-function createCallback(callback) {
-  return typeof callback === 'function' ? callback : logError;
-}
-
 function getMissingParams(params, required) {
   var missing = [];
 
@@ -85,7 +75,6 @@ function createAPIRequest(parameters, callback) {
   var req, body, missingParams;
   var params = parameters.params;
   var options = utils.extend({}, parameters.options);
-  var missingCallback;
   // If the params are not present, and callback was passed instead,
   // use params as the callback and create empty params.
   if (typeof params === 'function') {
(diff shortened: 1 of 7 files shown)


lib/googleapis.js
```diff
@@ -65,6 +67,80 @@ GoogleApis.prototype.addAPIs = function(apis) {
...
+  reqs.forEach(function(request, pos) {
+    //TODO check if vital components of a request exist
+
+    var body = [];
+    var path = request.url.replace('https://www.googleapis.com', '');
+
+    //if there is a query string, append it to the end of the path
+    if (request.useQuerystring && request.qs) {
+      var firstTime = true;
+      for (var key in request.qs) {
```
ryanseys: you could probably use a module like `qs` to handle this for you.

lib/googleapis.js
```diff
@@ -65,6 +67,80 @@ GoogleApis.prototype.addAPIs = function(apis) {
...
+ * Makes a batch request to google and receives an object with 
+ * E.g. googleapis.drive and googleapis.datastore
+ *
+ * @param {Object} params Params used to make a batch request. Must contain reqs
+ * @param  {Function} callback   Callback when batch request is finished or error found
+ */
+GoogleApis.prototype.batch = function(params, callback) {
+  var multipart = [],
+      rootUrl = 'https://www.googleapis.com';
+      //rootUrl = 'http://localhost:8080';
```
ryanseys: Remove

lib/googleapis.js
```diff
@@ -65,6 +67,80 @@ GoogleApis.prototype.addAPIs = function(apis) {
...
+  if (typeof params === 'object') {
+    reqs = params.reqs;
+    authClient = params.auth;
+  }
+  if (!Array.isArray(reqs)) {
+    return callback(new Error('reqs should be an array of requests.'));
+  }
+  
+  reqs.forEach(function(request, pos) {
+    //TODO check if vital components of a request exist
```
ryanseys: Don't leave TODOs in here. Not sure what this means.

lib/googleapis.js
```diff
@@ -65,6 +67,80 @@ GoogleApis.prototype.addAPIs = function(apis) {
...
+    multipart.push({
+      'Content-Type': 'application/http',
+      'Content-ID': pos,
+      body: body.join('\r\n')
+    });
+  });
+
+  return new BatchTransporter(authClient ? authClient : new DefaultTransporter()).request({
+    method: 'POST',
+    uri: rootUrl + '/batch',
```
ryanseys: So this is just https://www.googleapis.com/batch ?

lib/apirequest.js
```diff
@@ -85,10 +85,11 @@ function createAPIRequest(parameters, callback) {
   var req, body, missingParams;
   var params = parameters.params;
   var options = utils.extend({}, parameters.options);
-
+  var missingCallback;
   // If the params are not present, and callback was passed instead,
   // use params as the callback and create empty params.
   if (typeof params === 'function') {
+    missingCallback = false;
```
ryanseys: No need for this if you have `missingCallback = !callback;` later.

lib/apirequest.js
```diff
@@ -184,6 +186,10 @@ function createAPIRequest(parameters, callback) {
   delete options.auth; // is overridden by our auth code
   delete options.params; // We handle params ourselves and Request does not recognise 'params'
 
+  if (missingCallback) {
```
ryanseys: Just use !callback I'd say.

lib/apirequest.js
```diff
@@ -184,6 +186,10 @@ function createAPIRequest(parameters, callback) {
...
+    return options;
```
ryanseys: Will you also need to return the `body` potentially?

lib/batchtransporter.js
```diff
@@ -0,0 +1,65 @@
...
+
+    if (err || !body) {
+      return opt_callback && opt_callback(err, body, res);
+    }
+
+    if (res.headers['content-type'].indexOf('multipart/mixed') < 0) {
+      return opt_callback && opt_callback(new Error('Unexpected response type.'));
+    }
+
+    var boundary_regex = new RegExp('boundary=(.*)$', 'm');
```
ryanseys: Some comments here would be great

lib/batchtransporter.js
```diff
@@ -0,0 +1,65 @@
...
+    responses.forEach(function(response) {
+      var reduced_string = response.substr(response.indexOf('HTTP'));
+      var resp_obj = http_parser.parseResponse(reduced_string);
+
+      try {
+        var body = JSON.parse(resp_obj.body);
+        batch_response.push(body);
+      }
+      catch (err) {
+        //do nothing..?
```
ryanseys: Return an error in the callback here if possible. Else do nothing.

lib/googleapis.js
```diff
@@ -65,6 +67,80 @@ GoogleApis.prototype.addAPIs = function(apis) {
...
+
+    //request methods + path
+    body.push(request.method + ' ' + path);
+    //request headers
+    if (request.json) {
+      body.push('content-type: application/json');
+    }
+    body.push('');
+
+
```
ryanseys: Remove second new line

ryanseys:
Please add tests to cover this functionality.

lib/apirequest.js
ahageali: Was originally going to do that but the createCallback function guarantees that !callback will always evaluate to false

lib/apirequest.js
ahageali: Most likely but I'll need to find an API i can play with that uses body before figuring out how to incorporate it

lib/apirequest.js
ryanseys: You're right. If no callback is given then it's no longer an error so we can get rid of createCallback and logError function in this file.

lib/apirequest.js
ahageali: Regarding this, I've found that most scenarios that require the `body` variable causes the request to be a multipart request. 
For example, `drive.insert({media: {body: readStream}, uploadType: 'multipart', resource: {title: 'Test'}})` will cause a multipart request to be formed in the current API. If I were to implement the ability to support this in batch requests, it would require a lot of string manipulation which I'm unsure is suitable for the quality of API, as I would have to generate the multipart string into the request manually. 
Supporting a request that has the `body` variable, while not being multipart is pretty easy to implement though. 
Should I attempt to implement support for multipart requests inside batch requests?

ahageali:
Implemented support for streams as the request body and also multipart requests.
Tried to test using the upload API but found out (much) later that uploads are not supported through batch requests, so I'm not sure if any API actually exists that allows uploads while still being supported by batch requests.

I've also implemented some tests, though I'm not sure if they cover the functionality properly as I'm actually pretty new to node and haven't gotten into testing yet. Is it alright if the tests make real API calls, as emulating the requests/responses is tough due to the unique nature of the batch requests / responses.

package.json
```diff
@@ -35,6 +35,10 @@
   "dependencies": {
     "async": "~0.9.0",
     "gapitoken": "~0.1.2",
+    "http-string-parser": "0.0.4",
+    "memorystream": "^0.3.1",
+    "multipart-stream": "^1.0.0",
```
ryanseys: This module has a bug that fills the entire contents of the stream into memory, we previously used it and now have ripped it out of the library. I don't believe you'll need any of this stream functionality anyway because uploads are not supported by the batch API.

ahageali:
I have removed both `multipart-stream` and `memorystream` modules, and gotten rid of any support in the `batch` method for streams.

test/test.batch.js
```diff
@@ -0,0 +1,42 @@
...
+    var reqs = [];
+    reqs.push(drive.files.list({q: 'hello'}));
+    reqs.push(drive.files.get({fileId: '123abc'}));
+    googleapis.batch({reqs: reqs}, function(err, result) {
+      assert(err);
+      assert.equal(err.message, 'Forbidden');
+      assert(!result);
+      scope.done();
+      done();
+    });
```
bentona: How about a test to verify that each request is executed?

blackrabbit99:
@ryanseys  so what is your politics with batches requests? Is it going to add into node js client or no.
I've several thousands requests and it works awful. Or do you recommend to use third party libs or maybe this branch as temporary decision

ryanseys:
I no longer make big decisions with regard to this library, sorry. Please direct all future questions to @tbetbetbe. Thank you.

jokesterfr:
Hi, any news from this? Any plan to add this kind of feature in the project?
Thank you

blackrabbit99:
Looks like google team decided not support batch activity into their nodeJS library due to some reason.

I found it as a big issue for my project so make some part of batch api.
Take a look https://github.com/blackrabbit99/google-drive-batch-requests

It works with drive api now. It's not so hard to implement batch support for others api types based on http stack.

raz3r:
Is there any reason why this library doesn't support batch requests? Some APIs such as Directory API are quite useless without batch support...

JustinBeckwith:

grant:
Note: It also looks like global HTTP batch endpoints (www.googleapis.com/batch) will cease to work on March 25, 2019 anyways.

https://developers.googleblog.com/2018/03/discontinuing-support-for-json-rpc-and.html

PR #380, opened 2015-03-03
Change extend util to check hasOwnProperty

stephen:
The lack of a `hasOwnProperty` check can accidentally cause functions on the prototype to bubble up to being a parameter in an API request.

For example... defining a polyfill:
```
Object.prototype.entries = function* entries(obj) {
  for (let key of Object.keys(obj)) {
    yield [key, obj[key]];
  }
};
```

will cause an extra `entries` param to exist in a request in the query string, e.g. `https://www.googleapis.com:443'/calendar/v3/calendars/stephen%40stephenwan.net/events?entries=`.

### DIFF
--- a/test/test.utils.js
+++ b/test/test.utils.js
@@ -1,0 +1,45 @@
+/**
+ * Copyright 2014 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+var assert = require('assert');
+var utils = require('../lib/utils.js');
+
+describe('Utils', function() {
+
+  it('should not extend funtions on the prototype', function() {
+
+    var object = {
+      property: 'test'
+    };
+
+    /* jshint ignore:start */
+    Object.prototype.entries = function() {};
+    /* jshint ignore:end */
+
+    var newObject = utils.extend({}, object);
+
+    assert.equal('entries' in newObject, true);
+    assert.equal(newObject.hasOwnProperty('entries'), false);
+
+    assert.equal('property' in newObject, true);
+    assert.equal(newObject.hasOwnProperty('property'), true);
+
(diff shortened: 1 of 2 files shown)


ryanseys:
Please add a test.

stephen:
@ryanseys - Added a test that fails before the changes to `extends`, but passes currently. Satisfactory? Just checks that a prototyped function doesn't get copied over.

ryanseys:
Ahaha, jslint won't even let the build pass with it because extending Object is so bad. I guess we'll have to the scrap the test. I would highly suggest you don't extend native objects like that and would go as far to say that it's really not our fault if you do that and our code breaks because of it, but because this change is relatively straightforward and using hasOwnProperty is sort of "standard practice" in JavaScript due to limitations of `for .. in ..` I'll let it slide. God's speed [`for .. of ..`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/for...of)

stephen:
@ryanseys - I added a jshint ignore for the one line, so we can keep the test around.

Generally agree with not extending native objects, but I believe `Object.entries` is (was?) being considered at some point for ES7 support, and is fairly useful - https://esdiscuss.org/topic/object-entries-object-values.

test/test.utils.js
```diff
@@ -0,0 +1,43 @@
...
+    Object.prototype.entries = function() {};
+    /* jshint ignore:end */
+
+    var newObject = utils.extend({}, object);
+
+    assert.equal('entries' in newObject, true);
+    assert.equal(newObject.hasOwnProperty('entries'), false);
+
+    assert.equal('property' in newObject, true);
+    assert.equal(newObject.hasOwnProperty('property'), true);
```
ryanseys: Can you remove the property after the asserts so we don't poison the globals across all the tests?

Code:
``` js
delete Object.prototype.extends;
```

stephen: Will do

ryanseys:
Cool :) Thanks! Any way you could finally squash into 1 commit? Sorry for the troubles and thanks for your PR.

stephen:
No worries - was just in the middle of rebasing actually. :smile_cat:

ryanseys:
Thanks! Released in [`v1.1.5`](https://github.com/google/google-api-nodejs-client/releases/tag/v1.1.5)!

Alaneor:
I shall only add here as a tip that extending native objects **this way** is indeed considered a terrible practice exactly because of the side-effects you observed.

If you really need to do so, you must do it "properly", making sure your extensions are not in an enumerable property.

```
Object.defineProperty(Array.prototype, 'entries',
  { enumerable: false // default if omitted
  , value: function entries () { /* logic here */ }
  }
)
```

stephen:
@Alaneor - thanks for the tip. Realized that this was also causing a problem in [request](https://github.com/request/request) as well, and ended up switching to `defineProperty`.

PR #393, opened 2015-03-16
Fix for NodeJS Pub/Sub API

alessandro-aglietti:
On the v1beta2.js file all urls and REST path are wrong :-)

hope this help

### DIFF
--- a/apis/pubsub/v1beta2.js
+++ b/apis/pubsub/v1beta2.js
@@ -56,7 +56,7 @@ function Pubsub(options) {
       acknowledge: function(params, callback) {
         var parameters = {
           options: {
-            url: 'https://www.googleapis.com/v1beta2/{+subscription}:acknowledge',
+            url: 'https://pubsub.googleapis.com/v1beta2/' + params['subscription']  + ':acknowledge',
             method: 'POST'
           },
           params: params,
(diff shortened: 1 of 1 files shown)


ryanseys:
Thanks for bringing this issue to our attention. I've updated the generator/parser to consider the baseUrl in the generation process. You can see my fix here: https://github.com/google/google-api-nodejs-client/pull/394

ryanseys:
The fix has been released on npm as `v2.0.2`! Enjoy! :)

alessandro-aglietti:
The fix is ok but only for the domain issue.

There are also problems on the substitution for the placeholders in URLS, ie:
https://pubsub.googleapis.com/v1beta2/{subscription}:acknowledge

when used doesn't use the params.subscription

seem that the pathParams: ['subscription'] isn't used.

problems are in the googleapis/apis/pubsub/v1beta2.js file.

From my pull request you can see how I rapidly fix.

Hope this help

ryanseys:
Can you provide a code example that doesn't work? If you specify a
subscription parameter to the request it should properly be specified in
the url when making the request. Try it out and let me know if you can't
get it to work.

alessandro-aglietti:
Sorry it's works but require that on every projects we do
npm install --save googleapis
cd node_modules/googleapis
npm install
node scripts/generate --base-url "https://pubsub.googleapis.com"

any idea?

ryanseys:
Install the latest version 2.0.2 and you shouldn't have to do that.

PR #394, opened 2015-03-16
Fix endpoints for APIs by using baseUrl

ryanseys:
Fixes Pubsub and other endpoints that use a different baseUrl.


Alaneor:
It might be worth mentioning in the release notes that the `--base-url` switch has been removed to support this. I had my doubts about its usefulness but it is a breaking change, nonetheless.:)

ryanseys:
You're right, I should add that. Base url now will need to be specified in
the discovery doc if they want to generate a custom one.

PR #412, opened 2015-04-09
regenerate apis

Alex0007:


ryanseys:
Sorry, just got around to regenerating them. We need an automated way to do this :)

PR #420, opened 2015-04-26
Handle error in URL shortener example

adamcohen:
When I first came across the example in the Readme, I tried blindly copy/pasting it into a new sample js file and was surprised to see it failed with the following error:

`TypeError: Cannot read property 'longUrl' of null`

I was just about to submit a bug report, but I realized after adding an error check to the sample code that it was failing because no authorization credentials were provided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,11 @@ var params = { shortUrl: 'http://goo.gl/xKbRu3' };
 
 // get the long url of a shortened url
 urlshortener.url.get(params, function (err, response) {
-  console.log('Long url is', response.longUrl);
+  if (err) {
+    console.log('Encountered error', err);
+  } else {
+    console.log('Long url is', response.longUrl);
+  }
 });
 ```
 


README.md
```diff
@@ -56,7 +56,11 @@ var params = { shortUrl: 'http://goo.gl/xKbRu3' };
 
 // get the long url of a shortened url
 urlshortener.url.get(params, function (err, response) {
-  console.log('Long url is', response.longUrl);
+  if (err) {
+    console.log('Encountered error', err);
```
ryanseys: Add some spaces at the end of these strings perhaps?

adamcohen: It seems that `console.log` already adds spaces between the first and second arguments. If I add another space at the end of the string literals, I end up with two spaces between the string and the second argument.  At least this is the behaviour with node 0.12.2

Alaneor: Confirmed on Node.js 0.10, 0.12 - space is added automatically for each argument to `console.log`.

ryanseys: Oh yeah, I was thinking about `+` not `,`

PR #435, opened 2015-06-10
Avoid url-encoding path parameters.

TamalSaha:
GAPI node.js client currently url-encodes path parameters. This causes problem when path parameters contain '/'. This pull request stops url encoding path parameters. This makes node.js client consistent with GAPI Go client.

### DIFF
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -52,30 +52,6 @@ function getMissingParams(params, required) {
 }
 
 /**
- * Take a url path template and replace placeholders with data from params,
- * i.e. parsePath('/my/{value}', { value: 'something' }); // /my/something
- * @param  {String} path   String with template placeholders in curly braces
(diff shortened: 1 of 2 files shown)


ryanseys:
Hi Tamal, this change broke the tests (see the travis ci build for the associated errors). Please update the tests so that they continue to pass. Also, this is a breaking change so this will require a new major version, 3.0.0. I do not maintain the project anymore past regenerating the apis and merging trivial non-breaking changes that require patch-level increases in the version so someone else will need to approve this change before it can be released. I'm sorry for any inconvenience.

TamalSaha:
@ryanseys, all broken tests related to my change are fixed.

ilanle:
unsubscribe

tbetbetbe:
- All the tests should be green before this gets merged, so sorry but there will be a bit of a delay while the another PR fixing the current tests is added.

PR #443, opened 2015-06-23
Update README.md

rainabba:
Remove last parm (email address) from google.auth.JWT() call. It causes "Error authorizing with JWT [Error: unauthorized_client]" and nothing here explains why so I believe it's an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,8 +244,7 @@ var jwtClient = new google.auth.JWT(
   'serviceaccount@email.com',
   '/path/to/key.pem',
   null,
-  [scope1, scope2],
-  'bar@subjectaccount.com');
+  [scope1, scope2]);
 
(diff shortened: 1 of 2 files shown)


rainabba:
It looks as if my changes haven't taken which is for the better. The issue comes down to this: If you're using an API that requires impersonation, you must provide an authorized account. If you don't need impersonation, you can omit that parameter.

If you do need impersonation for a service account, visit the following as a domain admin to authorize a given service account for one or more scopes: https://admin.google.com/AdminHome?chromeless=1#OGX:ManageOauthClients

I found that information very hard to come by because the need isn't clearly addressed and the provided error doesn't lead to the appropriate documentation. In my case, it was available (if outdated) at https://developers.google.com/admin-sdk/reports/v1/guides/delegation

PR #444, opened 2015-06-23
Update "requests" to 2.58.0

josephpage:
Changelog from 2.54.0 : https://github.com/request/request/blob/master/CHANGELOG.md


bentona:
It looks like there were some changes to OAuth in 2.58 which broke some tests https://github.com/request/request/pull/1634

PR #452, opened 2015-07-09
Avoid parameter name conflicts with resource/media/auth

sqrrrl:
Simple escape mechanism to allow parameter names that happen to conflict with names the client library uses. Appends "_" to parameters that would otherwise conflict. Currently only a few APIs exhibit this issue, notably cloud APIs with the 'resource' parameter.


tbetbetbe:
LGTM, @ryanseys PTAL
I think we want to release a new version as a consequence of this change.
I don't think it's a major version change, but it's certainly a minor one - could you add the bump in this PR please ?

ryanseys:
Can you add a little note to the README for this edge case? Otherwise, LGTM . @tbetbetbe can you release this?

PR #463, opened 2015-08-04
Restore --discovery-url functionality to scripts/generate

theophoric:
scripts/generate.js attempts to parse all arguments as api urls.  This breaks the existing functionality of being able to provide a `--discovery-url` flag.  This patch restores this functionality while keeping the new behavior of generating apis for each un-named argument.

```
➜  google-api-nodejs-client git:(master) node scripts/generate --discovery-url "https://my.discovery.json" 
/Users/phaedrus/reference/javascript/google-api-nodejs-client/scripts/generate.js:33
        throw err;
              ^
Error: Invalid URI "--discovery-url"
```
An issue still exists where the apis generated will not be included in apis/index.js, as this file is generated from the discovery url file.

### DIFF
--- a/scripts/generate.js
+++ b/scripts/generate.js
@@ -21,10 +21,12 @@ var rimraf = require('rimraf');
 var path = require('path');
 var debug = false;
 
-var args = process.argv.slice(2);
+var argv = require('minimist')(process.argv.slice(2));
 
 // constructors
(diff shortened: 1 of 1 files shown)


jmdobry:
@theophoric Can you please verify that you have signed the CLA? https://developers.google.com/open-source/cla/individual

PR #485, opened 2015-11-07
change help link to google api keys

schmijos:
With the new console also the help changes.


jmdobry:
Both the old and new links don't work anywhere, so I went ahead and changed it to https://github.com/google/google-api-nodejs-client#using-api-keys

Thanks though! Sorry it took so long to get to this.

PR #488, opened 2015-11-11
Added support for building an API from a file.

jmdobry:
Sometimes, all you have is a file-based discovery doc. This PR makes those times easier to cope with.

### DIFF
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -23,6 +23,7 @@ var beautify = require('js-beautify').js_beautify;
 var path = require('path');
 var mkdirp = require('mkdirp');
 var fs = require('fs');
+var url = require('url');
 var argv = require('minimist')(process.argv.slice(2));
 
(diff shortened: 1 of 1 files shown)


waprin:
+1

While I would still prefer to see #488 done in addition to this, this can be used to solve the same problem by downloading the discovery doc and running this.

waprin:
@jmdobry I think some docs on how to use it would be useful. I generated the API ok but in order for it to have the 'google' object bound to it, it needs the requireAPI() function from apis/index.js which is not a public function. Currently the way I'm getting it to work is to copy the generated API into the directory and editing apis/index.js, but that means editing my node_modules/ which is not something I want to do. Is there a way you can use the generated API without editing it into this client library directly?

jmdobry:
Merging so I can get the code from my fork into the main fork, where I can continue working on it.

PR #499, opened 2015-12-21
path segments should be url encoded, closes #476

### LINKED ISSUE #476: 'Not Found' for searchanalytics.query, must encode siteUrl
Once auth'd up, I run

``` javascript
 var params = {
   auth: oauth2Client,
   siteUrl: 'http://my-domain.com/',
   resource: {
      startDate: '2015-09-01',
      endDate: '2015-09-30',
      dimensions: ['page']
   }
 };

  webmasters.searchanalytics.query(params, function(err, res) {
    if(err) console.log("err:", err);
    if(res) console.log("res:", res);
  });
```

Output gives:

```
res: 'Not Found'
```

All other API calls work, such as listing sites or sitemaps, and I definitely have access to the siteUrl specified. What is going wrong here?

ssuda:
path segments should be url encoded, please look at the following url
http://blog.lunatech.com/2009/02/03/what-every-web-developer-must-know-about-url-encoding

### DIFF
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -51,6 +51,32 @@ function getMissingParams(params, required) {
   return missing.length > 0 ? missing : null;
 }
 
+// Escaping path segement, please look at: http://blog.lunatech.com/2009/02/03/what-every-web-developer-must-know-about-url-encoding
+var pathSegmentEscapeChars = {
+  '#': '%23',
+  '/': '%2F',
+  '?': '%3F',
+  '[': '%5B',
+  ']': '%5D',
+  '%': '%25',
+  '<': '%3C',
+  '>': '%3E',
+  '\\': '%5C',
+  '^': '%5E',
+  '`': '%60',
+  '{': '%7B',
+  '|': '%7C',
+  '}': '%7D'
+};
+
+var pathSegmentEscapeCharsRegex = /[#\/\?\[\]%<>\\\^`,{\|}]/g;
+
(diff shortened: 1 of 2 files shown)


lib/apirequest.js
```diff
@@ -51,6 +51,32 @@ function getMissingParams(params, required) {
   return missing.length > 0 ? missing : null;
 }
 
+// Escaping path segement, please look at: http://blog.lunatech.com/2009/02/03/what-every-web-developer-must-know-about-url-encoding
+var pathSegmentEscapeChars = {
+  '#': '%23',
```
jmdobry: @ssuda What did you reference when you implemented this? The blog post you linked to does not cover the more technical aspects of safely encoding the path params component.

lib/apirequest.js
```diff
@@ -51,6 +51,32 @@ function getMissingParams(params, required) {
...
+  '>': '%3E',
+  '\\': '%5C',
+  '^': '%5E',
+  '`': '%60',
+  '{': '%7B',
+  '|': '%7C',
+  '}': '%7D'
+};
+
+var pathSegmentEscapeCharsRegex = /[#\/\?\[\]%<>\\\^`,{\|}]/g;
```
jmdobry: Where did this regular expression come from?

ssuda: These are reserved characters. Look at https://en.wikipedia.org/wiki/Percent-encoding

JustinBeckwith:

PR #511, opened 2016-01-30
fix linting error

MidnightWonderer:
Nobody seems to care about linting error.
Maybe they are busy doing more meaningful stuff.
Let me fix the linting for you then.


ryanseys:
Thanks!

PR #527, opened 2016-03-05
add glass mirror api sample

JustinBeckwith:


JustinBeckwith:
@jmdobry want to know what you think of the local server + oauth2 dance here.  In the Python samples the web server / oauth stuff is built in a single lib, maybe we do the same?

jmdobry:
Possibly, though that's beyond the scope of this PR.

Can you squash into 1 commit?

JustinBeckwith:
squashed

PR #550, opened 2016-04-11
Removed hard-coded apis directory file 

jmdobry:
- [x] `npm test` succeeds

This change makes it so you can add a new API to the client library without modifying `apis/index.js` (which has been deleted). This makes it much easier to work with APIs that are not in the main discovery service. You should no longer have to vendor this library into your project to work with unlisted APIs. You could just do the following:

```
npm i --save google-apis
```

Then modify your project's `package.json` file to include:

```json
"scripts": {
  "postinstall": "cd node_modules/google-apis/ && node scripts/generate <your-discovery-doc-or-url>"
}
```


JustinBeckwith:
LGTM

Issue #416, opened 2015-04-19
App Engine Cloud Endpoints support


juriwiens:
Hi,

In the 0.x versions I was able to use this client library with my App Engine Cloud Endpoints via a custom discovery like this:

```javascript
googleapis
    .discover(myEndpointsApiName, myEndpointsApiVersion, {baseDiscoveryUrl: "https://myapp.appspot.com/_ah/api/discovery/v1/apis/"})
```

Does a solution exist to interact with Cloud Endpoints in 2.x?

SpainTrain:
:+1: Any news on this?

ronbeltran:
I also wanted to know how to do this in version 2.x. If not i will be using version 0.8 for custom discovery of apis.

Gooseus:
Any sort of official response to this or a workaround to access Cloud Endpoints via node would be greatly appreciated.

erlichmen:
I wrote the original patch to support Cloud Endpoints with this package and I'm heartbroken it didn't make it to the 1.x and 2.x releases.

cyrus007:
Are the 0.x versions also broken, by which I mean google returns "Error 400 - Bad Request" for me ?

lostpebble:
Is this still not working with this library? Migrating from Java endpoints to a node setup at the moment and would like to use Endpoints for communication between my node and java modules. Searching around I haven't found any solutions for using App Engine Endpoints with Nodejs as of yet...

jmdobry:
Should be fixed by #555 

As of 5.2.0 you should be able to do:

```js
var google = require('googleapis');
google.discover('https://myapp.appspot.com/_ah/api/discovery/v1/apis', function (err, apis) {
  var myapi = google.myapi('v3');
  // use myapi
});
```

If anyone wants to try it, tell me how it goes.

PR #555, opened 2016-04-21
Added "load" method for discovering an API at runtime.

jmdobry:
Addresses:

- #416

What this PR adds:

Let's say that the latest `googleapis` in NPM supports `someapi` `v1`, you would do:

```js
var google = require('googleapis');
var someapi = google.someapi('v1');
```

But, what if `v2` of `someapi` is available, but the discovery doc isn't available in the main discovery index yet, or maybe it is and a new release of `googleapis` hasn't been generated yet. You can now do this:

```js
var google = require('googleapis');
google.load('http://domain.com/path/to/someapi/v2/discovery.json', function (err, someapi) {
  // use someapi
});
```

It also works for Cloud Endpoints v1:

```js
var google = require('googleapis');
google.discover('https://myapp.appspot.com/_ah/api/discovery/v1/apis', function (err, apis) {
  var myapi = google.myapi('v3');
  // use myapi
});
```

or

```js
var google = require('googleapis');
google.discoverAPI('https://myapp.appspot.com/_ah/api/discovery/v1/apis/myapi/v3/rest', function (err, myapi) {
  // use myapi
});
```

### DIFF
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -101,7 +101,7 @@ function createAPIRequest(parameters, callback) {
       delete params[key];
     }
   });
-  
+
   // Normalize callback
   callback = createCallback(callback);
 
--- a/lib/generator.js
+++ b/lib/generator.js
@@ -24,6 +24,8 @@ var path = require('path');
 var mkdirp = require('mkdirp');
 var fs = require('fs');
 var url = require('url');
+var util = require('util');
+var createAPIRequest = require('./apirequest');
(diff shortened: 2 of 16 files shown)


JustinBeckwith:
Could you include an example with the other samples?  Just so folks know how to use it?  Other than that, LGTM

jmdobry:
The generated jsdoc will include sample usage examples. The bigger sample will be in nodejs-docs-samples.

Issue #523, opened 2016-03-04
Add simultaneous promise and callback support


jmdobry:

GitTom:
+1

For example, see how they added promise support to AWS node library:  
https://blogs.aws.amazon.com/javascript/post/Tx3BZ2DC4XARUGG/Support-for-Promises-in-the-SDK

lirbank:
See also MongoDBs node driver. It's very nicely supports both callbacks and promises in the same driver.
http://mongodb.github.io/node-mongodb-native/2.2/api/

thepatrick:
It's incredibly confusing that the "JavaScript client library" [supports promises](https://developers.google.com/api-client-library/javascript/features/promises) but this library doesn't.

How can progress be made on this? We have `async`/`await` now, and those work brilliantly with libraries with promise support.

BerndErnst:
Seems like Google doesn't love NodeJs/Javascript 💔💔💔

simonh1000:
Docs recommend using https://developers.google.com/api-client-library/javascript/features/promises but I cannot find how to use them with this library

Mozcatel:
+1 
Seems we cant use the full asynchronous control Node allows us on this library, either Promises or async/await

machineghost:
So this ticket is nearly two years old and hasn't had any dev response in that entire time ... does that mean it's dead?

JustinBeckwith:
Nope, just slow moving :)  This project is largely maintained by volunteers on free time, so sometimes stuff like this can take a while.  We still very much want to do it, but a few things need to happen first. 

In the meantime - you can use tools like [pify](https://github.com/sindresorhus/pify) today to promisify the API pretty easily.  

We also love PRs :)

machineghost:
>We also love PRs :)

As I said I'd be happy to submit one, but only if it will actually be used :) I certainly wouldn't want to go through and `promisify` everything if the library's maintainers don't want that.

JustinBeckwith:
Apologies!  I did miss the edit.  I don't think we want to run through and `util.promisfy` everything :) . We're using TypeScript, which lets us write our code with promises or Async/Await, and then transpile down to es5 so we can support node 4 and 6.  I believe `util.promisfy` is only in Node 8, so that's sadly off the table.  That's were (if we need to, which we shouldn't) we could use pify.

[...]

Issue #197, opened 2014-07-07
Promise support


ryanseys:
Suggested by [this tweet](https://twitter.com/rayshan/status/486177042811998208) by @rayshan. Starting the conversation here about promise support.

eyy:
@ryanseys, might I ask to reconsider promise support? Now when we have native promises, and libraries like `co` leveraging generators, writing code like `var calendars = yield calendar.calendarList()` would be great.

Thanks.

ryanseys:
We aren't considering this in the near future but thanks for your feedback. I believe you can wrap the traditional callback structure with promises using other node libraries, so perhaps you could look into that if you need a Promises interface.

hess-g:
Bluebird's promisify works well against this lib. You can see an example [here](https://github.com/danthareja/node-google-apps-script/blob/master/lib/commands/upload.js#L40).

LegNeato:
AFAICT promisify doesn't work when doing JWT auth due to it not matching the node callback pattern. I had to construct a promise manually. I'm using ES7 async/await and arrow functions but should be trivial to do in vanilla ES6:

```js
// Create the promise so we can use await later.
const authPromise = new Promise((resolve, reject) => {
  jwtClient.authorize((err, tokens) => {
    if (err) {
      return reject(err);
    }
    // This is the part that doesn't happen when using promisify.
    jwtClient.setCredentials(tokens);
    resolve(jwtClient);
  });
});

// ...

// Auth as the JWT user. Throws if there is an error.
try {
  await authPromise;
} catch (err) {
  console.log(`ERROR: ${err.message}`);
  throw err;
}
```

ronenteva:
+1 for promise support

olalonde:
On a related note async/await is now implemented in v8: https://news.ycombinator.com/item?id=11710613 which in turn relies on promises... Libraries that don't support a promise based API will quickly run out of fashion.

AdamSLevy:
I think the fact that the JavaScript GAPI Library supports promises while the Node.js GAPI Library doesn't, is a point of confusion for new developers (like myself). Having a consistent JavaScript/Node.js interface would be nice.

liviubunda:
+1 for promise support

nhooey:
Is it possible use BlueBird's [`Promise.fromCallback()`](http://bluebirdjs.com/docs/api/promise.fromcallback.html) on this library?

ciscoo:
@nhooey Yes

```
[...]

PR #559, opened 2016-04-22
Implement backwards compatible promise support.

jmdobry:
This addresses #197 and #523 by adding backwards compatible promise support for API requests only. No changes were made to google-auth-library-nodejs. Promise support will need to be added separately to that library.

In a future release we may introduce breaking changes that improve promise support. This PR only requires a minor version bump. With this PR there are still edge cases where no `request` object is returned by the method call, in which case there is obviously no `request.promise` available. For example, when you try to make an API request but you haven't provided all the required params, no request object is returned, so no promise. =(

I'm not sure how useful this feature is until breaking changes can be made, unless you're sure you're passing in all required params and you can avoid the cases where no request object is returned by the method call.

I believe passing in an auth client with no tokens might be another case where no request object is returned...

Usage:

``` js
var req = urlshortener.url.insert({ resource: obj });

req.promise.spread((body, response) => {
  // ...
});
```

Example edge case with no promise available:

``` js
var req = adexchangebuyer.accounts.get({});
req // null, oh no!
```

have to pass a callback to get the error =(

``` js
var req = adexchangebuyer.accounts.get({}, function (err) {
  err; // Error: missing id
});
req // null, oh no!
```

### DIFF
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -17,6 +17,7 @@ var utils = require('./utils.js');
 var DefaultTransporter = require('./transporters.js');
 var stream = require('stream');
 var parseString = require('string-template');
+var Promise = require('bluebird');
 
 function isReadableStream (obj) {
   return obj instanceof stream.Stream &&
@@ -175,16 +176,37 @@ function createAPIRequest (parameters, callback) {
   delete options.auth; // is overridden by our auth code
   delete options.params; // We handle params ourselves and Request does not recognise 'params'
 
+  var resolve, reject;
+
+  // Promisify the results, but also call the node-style callback
+  var promise = new Promise(function () {
+    resolve = arguments[0];
+    reject = arguments[1];
+  }).asCallback(callback, { spread: true });
+
+  // Capture err and/or results
+  var promiseBack = function (err) {
+    if (err) {
+      return reject(err);
+    }
+    return resolve(Array.prototype.slice.call(arguments, 1));
+  };
+
   // create request (using authClient or otherwise and return request obj)
   if (authClient) {
-    req = authClient.request(options, callback);
+    req = authClient.request(options, promiseBack);
   } else {
-    req = new DefaultTransporter().request(options, callback);
+    req = new DefaultTransporter().request(options, promiseBack);
   }
 
-  if (body) {
-    body.pipe(req);
+  if (req) {
+    // Make promise available
+    req.promise = promise;
+    if (body) {
+      body.pipe(req);
+    }
   }
+
   return req;
 }
 
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
   },
   "dependencies": {
     "async": "~1.5.2",
+    "bluebird": "^3.3.5",
     "gapitoken": "~0.1.5",
     "google-auth-library": "~0.9.7",
     "request": "~2.72.0",
--- a/test/test.promises.js
+++ b/test/test.promises.js
@@ -1,0 +1,106 @@
+// Copyright 2013-2016, Google, Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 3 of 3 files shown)


JustinBeckwith:
I don't think we should introduce a promise API in this state (open for discussion of course if you want to do a hangout).  Hanging the promise off of the request object like this feels really weird. I get why you did it, but still.   Also, the inconsistency with the error handling is pretty rough.  

I'd rather make a full major version revision and break the API.  If I had to guess - most people probably aren't directly interacting with the returned request object from those APIs (it's not entirely documented).  If we made this a major breaking change, I'm guessing we could make it look like this?

```js
var promise = urlshortener.url.insert({ resource: obj }, function (err, result) {
  // there's a callback here for back compat, and because some people don't like promises
}).then((result) => {
  // promise folks can just use the promise returned from the method call (I know, breaking change)
}).error((error) => {
  // then we can use civilized error handling as well 
});
```

Thoughts?

jmdobry:
With a breaking change, yeah something like that. People still need a way to access the request object. I'm not sure the best way to expose it if the promise is what's being returned by the method call.

JustinBeckwith:
What's an example of something folks would do with the raw request object?  I can't off the top of my head think of anything, especially if I can't get at it until /after/ the request has been made.

jmdobry:
Currently, you're given the request object right away, synchronously. You
can listen to its stream events, or pipe it into something else.

ulion:
I see a lot library use the api call returned req object then .promise() to get the promise object to use. it's clean and do not affect exited code, and not a break change, why don't do that? and when will this be merged? @jmdobry

jmdobry:
Yeah, I could do that. I'll take another look at this later this week.

gunar:
Congrats on the the initiative! :balloon: 

1. I believe a breaking change is due, to make all methods have a Promise-based API. Let move on! :smiley: 

2. This is what I've been doing so far to get around callbacks:
  ```js
  const promisify = require('es6-promisify');
  promisify(sheets.spreadsheets.get)({
      includeGridData: true,
      spreadsheetId,
    }).then(doOtherStuff);
  ```

3. I wouldn't create a new dependency just for promises, though. node v6 already has native promises :+1:

  a. Optimally, we should let the user pass in their preferred promise implementation. `mongoose` does like this:
  ```js
  const mongoose = require('mongoose')
  // use native promises
  mongoose.Promise = global.Promise
  ```

Thank you!

jmdobry:
Yeah, I'm thinking about making a breaking change.

jimmywarting:
@jmdobry

instead of using spread (which is not in standard promise spec)
```javascript
req.promise.spread((body, response) => {
  // ...
});
```
You could use object destructuring instead
```javascript
req.promise.then(([body, response]) => {
  // ...
});

// This is probably what you also want to do when using async/await in v7
let [body, response] = await req.promise
```

jmdobry:
I looking into making a breaking change upstream in google-auth-library-nodejs

nhooey:
How can I get a `Promise` returned when calling `spreadsheets.values.get()`?

I'd like to be able to do this:
```javascript
let response = await google.sheets('v4').spreadsheets.values.get(request);
```

nhooey:
Is the issue fixed? It would help to mention why you're closing an issue before doing it.

jmdobry:
This should be fixed in the upstream google-auth-library-nodejs first. google/google-auth-library-nodejs#122 rewrites the HTTP layer, and that's the PR where we should pursue a breaking change to add Promise support.

florianajir:
What's new about Promise support for google-api-nodejs-client ?? For now it seems unable to make it worked ?!

tnguyen14:
@JustinBeckwith now that https://github.com/google/google-auth-library-nodejs/pull/182 is merged, is there any update on promise support?

JustinBeckwith:
Getting async/await support in google-auth-library helps, but there's still a fair bit of work to make this happen.  We totally want to do it, but in reality this is a volunteer project for most of us.  We're tracking this over in #523

PR #562, opened 2016-04-22
Hotfix. Fixes #560

### LINKED ISSUE #560: 5.2.0 is messed up - missing dependencies, missing files
After installing googleapis@5.2.0 as a dependency with npm and requiring it in my code:

 - Dependency "swig" is missing (so I installed it manually)
 - Dependency "js-beautify" is missing (I installed it manually)
 - The attempt to open file "./templates/api-endpoint.js" fails

That's as far as I went, then reverted to 5.1.0 which works just fine.

jmdobry:


JustinBeckwith:
What was the tl;dr root cause?  After the hotfix, I assume you're looking at tests for whatever happened.

jmdobry:
Yes, I will devise a test that installs `googleapis` with `NODE_ENV` set to production.

PR #612, opened 2016-08-05
Upgrade request dependency, fixes #608

### LINKED ISSUE #608: Security vulnerability in tough-cookie through request
`tough-cookie` has a ReDOS vulnerability that is fixed in 2.3.0. This repo depends on `request`, which has recently been updated to require the fixed version. See https://nodesecurity.io/advisories/130.

It would be great if this could be updated to support 2.74.0 of `request` and released.

Note that `google-auth-library` also requires `request`, so this will need be updated also, and the updated version used here also.

bengourley:
Fixes #608.

- [x] `npm test` succeeds

`tough-cookie`, a transitive dependency of this module via `request`, has a security vulnerability:

https://nodesecurity.io/advisories/130

`request` has already upgraded its dependency to fix the issue, so it's just a matter of bumping the version depended on here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "async": "~1.5.2",
     "gapitoken": "~0.1.5",
     "google-auth-library": "~0.9.7",
-    "request": "~2.72.0",
+    "request": "^2.74.0",
     "string-template": "~1.0.0",
     "url": "^0.11.0"
   },


bengourley:
I signed it!

jmdobry:
Thanks!

PR #622, opened 2016-08-23
Adding ability to override the endpoint URL

### LINKED ISSUE #621: Override API endpoint URL
With preventExtensions turned on, it is not possible to duck-punch the ability to override API endpoint URLs. We need to do this because we are proxying GA requests so they can be properly rate limited and accounted for. Ideally, one would be able to add a key called 'url' to the params object passed into apirequest.createAPIRequest(). If provided, it can override the existing value.

ianmaddox:
Fixes #621 (it's a good idea to open an issue first for discussion)
- [x] `npm test` succeeds
- [x] Approprate changes to readme/docs are included in PR

With preventExtensions turned on, it is not possible to duck-punch the ability to override API endpoint URLs. We need this because we are proxying GA requests so they can be properly rate limited and accounted for. This small change makes that possible while maintaining backwards compatibility.

### DIFF
--- a/lib/apirequest.js
+++ b/lib/apirequest.js
@@ -112,6 +112,11 @@ function createAPIRequest (parameters, callback) {
     return null;
   }
 
+  // Allow for proxied endpoints
+  if (params.url) {
+    options.url = params.url;
+  }
+
   // Parse urls
   if (options.url) {
     options.url = parseString(options.url, params);


ianmaddox:
I signed it!

jmdobry:
I have a feeling that's there a possibility that this might conflict with some API that actually allows a param called `url`.

What do you think about something like this?

```js
urlshortener.insert({
  resource: {
    longUrl: 'http://google.com/'
  }
}, {
  url: 'https://myurlshortenerproxy.com'
}, function (err, body) { /* ... */ });
```

jmdobry:
Do you think this PR has been obviated by #623?

LegNeato:
I believe with https://github.com/google/google-api-nodejs-client/pull/623 and https://github.com/google/google-api-nodejs-client/pull/756 you can override the whole URL or only the `rootUrl` portion (so you don't have to deal with all the service paths if they are unchanging) so this should be closed.

jmdobry:
Yep, sounds good.

PR #623, opened 2016-08-23
Add support for overriding request options.

### LINKED ISSUE #501: Drive download returning token JSON object
Hello, I'm having a problem downloading from the Drive using this API (v3). The data returned from the API is of the form: 

    {
      "access_token" : "HUGE_TOKEN_STRING",
      "token_type" : "Bearer",
      "expires_in" : 3600
    }

My code is roughly as follows (parts ommited due to confidentiality reasons):

      var buffer = new Buffer(0);
      drive.files.get({
          auth: auth,
          alt: 'media',
          fileId: id,
      })
      .on('error', function (err) {
        logger.error("Error " + err);
        callback(err, null);
      })
      .on('data', function (data) {
        logger.info("Got data!");
        buffer = Buffer.concat([buffer, data], buffer.length + data.length);
      })
      .on('end', function () {
        logger.info("End event!");
        callback(null, buffer);
      });

auth is an object created using a derivation of this example: https://developers.google.com/drive/v3/web/quickstart/nodejs#step_3_set_up_the_sample

Am I doing something wrong, or has the API changed somehow?

### LINKED ISSUE #618: Failed to export a file using drive API
I am using:
* googleapis@12.3.0
* node@4.3.2

I was trying to download a spreadsheet using Drive API v3 like this:
```js
drive.files.export({
    fileId: <some-valid-file-id>
    mimeType: 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
}, (err, buf) => {
    if (err) {
        console.log('Error:', err);
        return;
    }

    console.log('Received %d bytes', buf.length);
    fs.writeFileSync('./out.xlsx', buf);
});
```

Resulting file was corrupted. Same situation with 'application/pdf' - I could not open them.
After some investigation, I noticed that node-request module downloads data as string (decoded as utf8) by default unless you specify encoding option set to `null`. (what I am downloading here are binary files!)

```
// cited from - https://github.com/request/request
encoding - Encoding to be used on setEncoding of response data. If null, the body is returned as a Buffer. Anything else (including the default value of undefined) will be passed as the encoding parameter to toString() (meaning this is effectively utf8 by default). (Note: if you expect binary data, you should set encoding: null.)
```

By adding a line in drive.files.export() function like below, I was able to download the file successfully:
```js
// ./apis/drive/v3.js
    export: function (params, callback) {
      var parameters = {
        options: {
          url: 'https://www.googleapis.com/drive/v3/files/{fileId}/export',
          method: 'GET',
          encoding: null // <===== added line
        },
        params: params,
        requiredParams: ['fileId', 'mimeType'],
        pathParams: ['fileId'],
        context: self
      };
```

I am aware this file is generated with swig, and probably other methods (like drive.files.get with "alt" set to "media" for binary data) are suffering from the same problem, I do not have sufficient knowledge to propose most effective solution to this. I hope some one could come up with a good idea to this problem.

### LINKED ISSUE #621: Override API endpoint URL

jmdobry:
- [x] `npm test` succeeds

Fixes #501
Fixes #618
Fixes #621 

## Before

```js
drive.files.export({
  fileId: 'asxKJod9s79', // A Google Doc
  mimeType: 'application/pdf',
  encoding: null // This is ignored
}, , function (err, buffer) {
  // Oh no! buffer got incorrectly interpreted as a string!
});
```

## After

```js
drive.files.export({
  fileId: 'asxKJod9s79', // A Google Doc
  mimeType: 'application/pdf'
}, {
  encoding: null // Make sure we get the binary data
}, function (err, buffer) {
  // Nice, the buffer is actually a Buffer!
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,7 +500,7 @@ google.auth.getApplicationDefault(function (err, authClient) {
   var request = {
     projectId: '<your-project-id>',
     datasetId: '<your-dataset-id>',
-    
+
     // This is a "request-level" option
     auth: authClient
   };
@@ -512,6 +512,22 @@ google.auth.getApplicationDefault(function (err, authClient) {
       console.log(result);
     }
   });
+});
+```
+
+You can also override *request* options per request, such as `url`, `method`,
+and `encoding`.
+
+For example:
+
+```js
+drive.files.export({
+  fileId: 'asxKJod9s79', // A Google Doc
+  mimeType: 'application/pdf'
+}, {
+  encoding: null // Make sure we get the binary data
+}, function (err, buffer) {
+  // ...
 });
 ```
 
--- a/templates/api-endpoint.js
+++ b/templates/api-endpoint.js
@@ -19,6 +19,7 @@
 'use strict';
 
 var createAPIRequest = require('../../lib/apirequest');
+var utils = require('../../lib/utils');
 
 {% set Name = name|capitalize %}
 {% set Version = version|replace('\.', '_')|capitalize %}
--- a/templates/method-partial.js
+++ b/templates/method-partial.js
@@ -32,15 +32,22 @@
  * @param {object} params.resource Request body data
 {% endif -%}
 {% endif -%}
+ * @param {object} [options] Optionally override request options, such as `url`, `method`, and `encoding`.
  * @param {callback} callback The callback that handles the response.
  * @return {object} Request object
  */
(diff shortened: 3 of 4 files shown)


jmdobry:
@JustinBeckwith Looks good to you?

PR #715, opened 2017-03-10
18.0.0

jmdobry:


tcoffee-google:
In light of https://github.com/google/google-api-nodejs-client/pull/715/files#diff-354f30a63fb0907d4ad57269548329e3 should we update the "engines" specification in `package.json`?

jmdobry:
Probably, yeah.

jmdobry:
And also call it out in the breaking changes in the changelog

Issue #503, opened 2016-01-13
Typescript definition file


Bigous:
Could you include the d.ts file in this client distribution?

isman-usoh:
+1

panKt:
+1 :+1:

procrastinatos:
+1

christhomas:
yes absolutely! this project really needs a typescript definition file....

procrastinatos:
I took a stab at generating Typescript definitions based on the endpoint data from Google's discovery API. Disclaimer: I'm fairly new to all things Node & Typescript. Caveat emptor. Anyhow, typings available in the dist folder.

Bigous:
It seams that the new SALSA technology, typescript will be able to recognize things without .d.ts - [there is a pull request enable this things](https://github.com/Microsoft/TypeScript/pull/7075).

VamshiKrishnaAlladi:
+1

arondesign:
+1

n-sviridenko:
+1

danieloprado:
+1

masimplo:
Any news on this? Typings would greatly improve our experience working with this library.

abruneau:
+1

abruneau:
Since Typescript 2.1 you can import Untyped modules 
So you can do 
`import * as google from "googleapis";`

But for that, we need to update the `package.json` file and change main to `"main": "./lib/googleapis.js",`

JustinBeckwith:
I'm starting to work on this one.  This has to be done over several PRs, but the plan is:
- [ ] Perform a basic update the files to `*.ts` and wire up the compiler (#731)
- [ ] Fix syntax to move towards classes and es6 style imports
- [ ] Introduce tslint with relaxed configuration
- [ ] Add `noImplicitAny` to the `tsconfig.json` and apply required types
- [ ] Include the `.d.ts` file along with the npm module

Feedback on the approach welcomed.

gDelgado14:
hey @JustinBeckwith any update on this? You also don't need a `.d.ts` file since the code is written in typescript. The typedefs should be baked into the code itself. Otherwise the simple fix would have been to leave all of the code as it was and simply include a `.d.ts.` file to describe the structure of the package to the ts compiler. 

[Axios](https://github.com/axios/axios/blob/master/index.d.ts) is a good example of this approach.

gDelgado14:
[...]

PR #731, opened 2017-04-24
First step to TypeScript upgrade

JustinBeckwith:
First steps towards #503.  This simply renames all the files (including generated ones) to *.ts, and makes any changes necessary to get the build and tests to pass.  I am going to update #503 with the full list of steps we need to take here.  

- [x] `npm test` succeeds

I chose to hand modify apis/ to mollify the TypeScript compiler, and because I didn't want to complicate the PR with schema changes.


JustinBeckwith:
@matthewloring since @ofrobots is out, can you take a look?

package.json
```diff
@@ -77,16 +79,20 @@
...
-    "swig": "1.4.2"
+    "source-map-support": "^0.4.14",
+    "swig": "1.4.2",
+    "typescript": "^2.2.2"
   },
   "scripts": {
-    "lint": "semistandard \"**/*.js\"",
-    "generate-apis": "node scripts/generate",
+    "lint": "semistandard \"samples/**/*.js\"",
+    "generate-apis": "npm run build-tools && node scripts/generate.js",
```
jmdobry: Shouldn't `build-tools` run _after_ `scripts/generate.js`?

JustinBeckwith: Comically, no.  build-tools actually builds generate.ts into generate.js so it can be run by node.

jmdobry: does generate.js not produce more .ts files?

lukesneeringer:
> This simply renames all the files (including generated ones) to *.ts

This seems wrong to me unless the thing that generates the code has been updated to spit out `*.ts` files. Hand-edits to autogen code need to not happen.

package.json
JustinBeckwith: Yes :) `npm run build-tools` generates `generate.js`, and `npm run generate-apis` generates `apis/*.ts`, which `npm run build` compiles into js.  Sorry - I know this is a little confusing.

JustinBeckwith:
@lukesneeringer I updated the generators and templates to emit *.ts files, re-ran the generator, and ensured that it all worked.  The problem is that regenerating the ts files is non-deterministic.  The PR would include surface changes, and about 200k lines across all generated files to review.

It's really up to you guys.  I'm trying to limit the scope and improve the readability of the PR - but I'm happy to throw a watermelon over the fence :)

lukesneeringer:
If you updated the generator, that resolves my complaint.

lib/discovery.ts
```diff
@@ -59,13 +58,14 @@ function makeMethod (schema, method, context) {
       params: params,
       requiredParams: method.parameterOrder || [],
       pathParams: getPathParams(method.parameters),
-      context: context
+      context: context,
+      mediaUrl: null
```
matthewloring: Why is this option added? If this was done only for typescript maybe it should be made optional at the definition site.

test/test.clients.ts
```diff
@@ -131,10 +133,10 @@ describe('Clients', function () {
   it('should support default params', function (done) {
     var google = new googleapis.GoogleApis();
     var datastore = google.datastore({
-      version: 'v1beta2',
+      version: 'v1beta3',
```
matthewloring: Why were these logical changes included as well?

lib/discovery.ts
JustinBeckwith: It was done just for TypeScript, but this object literal is what formed the implicit type the compiler was checking against.  When I do a proper typing PR, I will move this to an interface with an optional parameter.

test/test.clients.ts
JustinBeckwith: Funny story!  The v1beta2 release of the datastore *doesn't exist anymore*.  These unit tests are currently failing in the master branch.  I switched to `v1beta3` which actually changed the API surface.  So ...  I had to switch the unit tests.

matthewloring: It feels like this change should be separated from the move to typescript but the owners of the repository would probably be the right people to comment on whether this is ok.

jmdobry: keep this change.

jmdobry:
I'm going to test this branch against our Node.js samples that use google-api-nodejs-client

jmdobry:
Tested against our googleapis samples in the nodejs-docs-samples repo.

PR #732, opened 2017-04-26
Arrow functions, let/const, and cleanup

JustinBeckwith:
This introduces let/const instead of var, moves to arrow functions in some cases, removes explicit 'use strict', and cleans up a few imports.


jmdobry:
Do you plan on switching to ES2015 classes, destructuring, etc. in a separate PR?

JustinBeckwith:
@jmdobry Yeah, a basic outline of my plan is in #503.  After this, next step is es6 style imports and classes.

PR #736, opened 2017-05-01
Update the generator to es6 imports, run generator

JustinBeckwith:
Just another step in the TypeScript #503 work


JustinBeckwith:
This is broke.  Will fix in a bit....

PR #742, opened 2017-05-10
Target es6, use Object.assign

JustinBeckwith:
Simple clean up PR that covers a few things:
- Changes TS config to target es6 instead of es5
- Uses `Object.assign` instead of the util lib in the auth library
- Switch a few modules over to es6 style imports


lukesneeringer:
Does the `import` statement work in Node 4? I thought it did not.

ofrobots:
@lukesneeringer note that this is in typescript now.

test/test.auth.ts
```diff
@@ -15,8 +15,9 @@ import * as assert from 'power-assert';
 import * as async from 'async';
 import * as nock from 'nock';
 import utils from './utils';
-let googleapis = require('../');
-let googleauth = require('../lib/googleauth.js');
+import googleauth from '../lib/googleauth';
+const googleapis = require('../');
```
jmdobry: Why can't `import` be used here?

JustinBeckwith: Right now googleapis.ts doesn't currently export an es6 module.  That change is very involved, and best kept separate from this one.

ofrobots:
Re: target es6: that wouldn't necessarily work with Node 4 as it doesn't have complete ES6 support.

<insert lament about TypeScript offering implementation levels rather than spec levels in their target options>

jmdobry:
A good example would be array destructing, can't use it in our samples :(

JustinBeckwith:
I am counting on the Travis tests that cover Node 4 here :)  Are you ok with me targeting es6?

ofrobots:
Given that the code coverage is at 79%, are we confident that we can trust the tests on this matter?

JustinBeckwith:
Some good news here is that the scripts/ dir, which contains the generator code, is by far the biggest culprit here:

image

I can work on getting better coverage for discovery.ts before moving on.

PR #757, opened 2017-06-06
Allow falsy params

saicheems:
This commit addresses two issues concerning falsy param values:

1. User confusion

   Given a method with a required parameter `foo`, if a request is made
   with the following object:

   ```
   {
     foo: 0
   }
   ```

   The following error is returned:

   ```Error: Missing required parameters: foo```

   This message is confusing - it's not obvious that the issue is that
   `foo` cannot be zero. This commit allows the above request to pass to
   the server, with the expectation that the server returns a more
   appropriate error message.

   The alternative solution, writing a special error message stating
   that falsy values are not allowed for required parameters, is
   complicated because of the second issue (below).

2. Required boolean params

   The `compute:v1` API contains a method,
   `compute.instances.setDiskAutoDelete`, which has a required boolean
   parameter `autoDelete`. If a request is made with the following
   object:

   ```
   {
     ... // Other required params.
     autoDelete: false
   }
   ```

   The following error is returned:

   ```Error: Missing required parameters: autoDelete```

   This is a bug in the client library, it should be possible to make a
   request with `autoDelete` set to `false`.

- [x] `npm test` succeeds
- [x] Approprate changes to readme/docs are included in PR

### DIFF
--- a/lib/apirequest.ts
+++ b/lib/apirequest.ts
@@ -38,7 +38,7 @@ function getMissingParams (params, required) {
 
   required.forEach(function (param) {
     // Is the required param in the params object?
-    if (!params[param]) {
+    if (!params.hasOwnProperty(param)) {
       missing.push(param);
     }
   });
--- a/lib/apirequest.ts
+++ b/lib/apirequest.ts
@@ -38,7 +38,7 @@ function getMissingParams (params, required) {
 
   required.forEach(function (param) {
     // Is the required param in the params object?
-    if (!params.hasOwnProperty(param)) {
+    if (params[param] === undefined) {
       missing.push(param);
     }
   });
--- a/test/test.query.ts
+++ b/test/test.query.ts
@@ -18,6 +18,7 @@ import utils from './utils';
 let googleapis = require('../');
 
 describe('Query params', () => {
+  let localCompute, remoteCompute;
   let localDrive, remoteDrive;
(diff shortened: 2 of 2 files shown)


saicheems:
Ah, sorry for not creating an issue before - please let me know if you'd like me to make one now for discussion.

I noticed this issue when running sample code against mocked APIs - I figured the easiest way to solve it was to drop the check for truthiness of parameters.

saicheems:
fyi @tcoffee-google @vchudnov-g

lib/apirequest.ts
```diff
@@ -38,7 +38,7 @@ function getMissingParams (params, required) {
 
   required.forEach(function (param) {
     // Is the required param in the params object?
-    if (!params[param]) {
+    if (!params.hasOwnProperty(param)) {
```
jmdobry: I would prefer this be changed to: `if (params[param] !== undefined) {`, which covers a few more edge cases.

saicheems: Did you mean `if (params[param] === undefined) {`?

Done, PTAL

saicheems:
Note that the test failure is an unrelated issue to this PR

lib/apirequest.ts
jmdobry: Yeah, that's what I meant

PR #767, opened 2017-06-28
Regen all APIs

saicheems:


saicheems:
1. Ran `npm run generate-apis`
2. Bumped the version in `package.json` to `19.1.0`
3. Added notes to `CHANGELOG.md`

Let me know if I need to do anything else.

saicheems:
Note that the build failure is due to `dataproc:v1alpha1`, which is returning a 404 for it's discovery doc. It's not actually an issue with the library.

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+##### 19.1.0 - 27 June 2017
```
jmdobry: Please bump to 20.0.0 instead, as there are breaking changes. Thanks!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "googleapis",
-  "version": "19.0.0",
+  "version": "19.1.0",
```
jmdobry: Please bump to 20.0.0 instead, as there are breaking changes. Thanks!

CHANGELOG.md
saicheems: Ah, right - done!

package.json
saicheems: Done

PR #773, opened 2017-07-07
Add notice on library maintenance

danoscarmike:


danoscarmike:
@jmdobry @lukesneeringer suggest we also edit the repo description to remove the "Google's officially supported " leader.

PR #790, opened 2017-08-14
refactor: get rid of googleauth internals

ofrobots:
googleauth.ts was reaching into the internals of google-auth-library. Most of it wasn't needed here.


JustinBeckwith:
CI issues aside, LGTM!

PR #792, opened 2017-08-15
build: improve npm scripts

ofrobots:


jmdobry:
LGTM, test failures are unrelated, it's still a problem with `container.googleapis.com format: rest version: v1alpha1` in the discovery service.

PR #798, opened 2017-08-17
Adding the 'Service account' keyword

steren:
Service Tokens are the solution to use when using Service accounts. Adding the string 'service account' will help people performing a text search on our README


steren:
This is a README change... how can Travis fail? How to re-run it?

PR #800, opened 2017-08-21
Warn on failure in the Discovery generation test

saicheems:
The test in `test.discover.ts` frequently fails because the Discovery service can return URLs to Discovery documents that 404. For the better part of the last 2 months, that test has failed because some service has a document URL which 404s, which makes automating updates difficult. This commit modifies that test to log a warning on failure, so it never fails the suite.

- [x] `npm test` succeeds
- [ ] Approprate changes to readme/docs are included in PR


saicheems:
Note: The alternative to dropping the test is to reconfigure the generator to ignore 404s and skip the initial clean-up step that deletes the apis/ directory, but I wasn't able to figure out how to do the former in a nice way.

saicheems:
That's a good idea, is there a way to change a test to emit a warning instead of causing a failure?

I haven't immediately found a way just Googling about Mocha

test/test.discover.ts
```diff
@@ -1,55 +0,0 @@
...
-    localApis.forEach((name) => {
-      assert(google[name]);
-      google[name] = undefined;
-    });
-
-    assert.equal(google.drive, undefined);
-
-    google.discover('https://www.googleapis.com/discovery/v1/apis', (err) => {
-      if (err) {
-        return done(err);
```
jmdobry: Yes, so you would keep the whole test file as-as, then change this one line from `return done(err);` to

```js
console.warn(err);
return done();
```

saicheems: Done, thanks for the explanation!

saicheems:
Friendly ping, lgty?

Issue #716, opened 2017-03-12
state not being sent


1mike12:
<your detailed question and code samples go here>
For some reason my state parameters are not being places into the final url on generateAuthUrl(). 

My client id, secrets, access type, scopes, and redirect_uri are all generated correctly, 

using this code: 

```javascript
let url = this.client.generateAuthUrl({
            // 'online' (default) or 'offline' (gets refresh_token)
            access_type: 'offline',
            // If you only need one scope you can pass it as a string
            scope: SCOPES,
            // Optional property that passes state parameters to redirect URI
            state: {foo: "bar", email: emailAddress, "test": "one"}
        });
```

I get 

```
https://accounts.google.com/o/oauth2/auth?access_type=offline&scope=https%3A%2F%2Fmail.google.com%2F&state=&response_type=code&client_id=xxx.apps.googleusercontent.com&redirect_uri=http%3A%2F%2Flocalhost%3A9000%2Foauth2callback%2Fgoogle
```

1mike12:
It may have something to do with my version of node, but the problem starts with querystring.stringify in oauth2client.js, line 125. 

In node 6.3.1, this leads into querystring.js line 210, where stringifyPrimitive(v) doesn't work for an object.

1mike12:
for now setting state as a Base64 string is an ok workaround

matkoklaic:
Just make sure to encode the state object passed to `generateAuthUrl` function:
`state: encodeURIComponent(JSON.stringify(state))`.

alfonsomga:
Stringifying and enconding state object did the trick as @matkoklaic mentioned.

This should be included in the documentation.

ace-n:
@alfonsomga Mind s

alexwohlbruck:
The docs aren't mentioning this at all, and actually give an example passing an object literal, which is misleading:

```
var url = oauth2Client.generateAuthUrl({
  // 'online' (default) or 'offline' (gets refresh_token)
  access_type: 'offline',

  // If you only need one scope you can pass it as a string
  scope: scopes,

  // Optional property that passes state parameters to redirect URI
  // state: { foo: 'bar' }
});
```

kiriti999:
I am facing the same issue. It doesn't redirect to the callback url, instead it just hangs the page there

JustinBeckwith:
Greetings folks!  Apologies for the troubles.  We've updated the readme to use a string:
[...]

PR #817, opened 2017-09-27
Fix wrong documentation for generateAuthUrl()

### LINKED ISSUE #794: generateAuthUrl state parameters issue
The function generateAuthUrl  is not accepting objects as state parameters, it only accepts strings/numbers. The doc example doesn't work:
```
var url = oauth2Client.generateAuthUrl({
  // 'online' (default) or 'offline' (gets refresh_token)
  access_type: 'offline',

  // If you only need one scope you can pass it as a string
  scope: scopes,

  // Optional property that passes state parameters to redirect URI
  // state: { foo: 'bar' } //does not work
     state: 'bar' //works
});
```
If you pass an object in the state, the callback will return just the code parameter. If you pass string/number, the callback will return the state and the code.

sohlex:
Hi everyone!
The `state` parameter in the auth URL should be a string, as documented here https://developers.google.com/identity/protocols/OAuth2WebServer .

In this README was stated that is possible to pass the `state` as an object literal, but this doesn't work. Changing the parameter to a string works like expected.

Fixes #794 #716

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ var url = oauth2Client.generateAuthUrl({
   scope: scopes,
 
   // Optional property that passes state parameters to redirect URI
-  // state: { foo: 'bar' }
+  // state: 'foo'
 });
 ```
 ##### IMPORTANT NOTE


sohlex:
I signed it!

bantini:
@sohlex Thanks for the PR.
@ofrobots @lukesneeringer @saicheems Do you think this PR is ready to be merged?

Issue #803, opened 2017-08-25
API broken due to compilation issue


niftylettuce:
Since `import` is not built in, the `googleapi` package is currently broken.

You cannot `require('googleapis/lib/apirequest')` for instance as you could before (you'd have to do `.default` after.

Perhaps add something similar to <https://github.com/59naga/babel-plugin-add-module-exports> and publish new version?

niftylettuce:
It appears v20 dropped support for this.  Can you add it back?  Node is not using `import` syntax yet.

niftylettuce:
Also, it looks like this package does not use default exports, but this package does?

<https://github.com/google/google-auth-library-nodejs/commit/62e801c4b78594d7bb3f801a04e5adcc198df30e>

Very confusing to have two different styles by Google for very closely related API libraries.

ofrobots:
I think this is due to how default exports work in TypeScript compiled code. A fix similar to https://github.com/google/google-auth-library-nodejs/commit/4f74dfafe86cdb2c5c36ffc29bdf0d1bafaff71b might be needed here.

niftylettuce:
This is still an issue.  See <https://stackoverflow.com/a/40357573>.

Please fix this asap, thank you!

ofrobots:
Fix being worked upon in #837.

JustinBeckwith:
Greetings folks!  I believe we resolved this issue by completely getting rid of default exports in a recent release.  Please take a look, and open a new issue if you're still having problems!

PR #837, opened 2017-11-02
803 broke compilation fix

bantini:
Fix for #803 broken compilation issue

### DIFF
--- a/test/test.auth.ts
+++ b/test/test.auth.ts
@@ -15,15 +15,16 @@ import * as assert from 'power-assert';
 import * as async from 'async';
 import * as nock from 'nock';
 import utils from './utils';
-const googleapis = require('../');
+import googleapis from '../lib/googleapis';
+const google = new googleapis();
 
 
 describe('JWT client', () => {
   it('should expose the default auth module', () => {
-    assert(googleapis.auth.getApplicationDefault);
+    assert(google.auth.getApplicationDefault);
   });
   it('should create a jwt through googleapis', () => {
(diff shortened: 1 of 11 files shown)


bantini:
CLA signed.

lib/googleapis.ts
```diff
@@ -211,4 +211,4 @@ GoogleApis.prototype.discoverAPI = function (path, options, callback) {
  * @module googleapis
  * @type {GoogleApis}
  */
-export = new GoogleApis();
+export default GoogleApis;
```
ofrobots: The Google TS style guide disallows default exports. This would be semver-major, but this change is semver major either way. /cc @lukesneeringer @JustinBeckwith

ofrobots: See also this comment about lack of consistency in default export usage: https://github.com/google/google-api-nodejs-client/issues/803#issuecomment-324934917. I think it is a fair criticism.

bantini: apirequest.ts and discovery.ts use the export default syntax as well. Do we need to convert these to a class based declaration like the one used in the auth library?

ofrobots: @bantini yes, it would be good to avoid default exports from those files too. However, those are not public APIs for this module, so you can fix them in a separate PR.

JustinBeckwith: What's the status of this?

bantini: With all the new changes, it would be easier to start again rather than repair

PR #856, opened 2017-11-13
BLOCKED: Fixes #741

### LINKED ISSUE #741: The library is not compatible with webpack
The library is not compatible with webpack

jmdobry:
Removes dynamic `require` statements. Generates an `index.ts` to eagerly load all APIs.

See #741 

BLOCKED: Need to run the generator, but it no longer has access to the code snippet fragments.


apis/acceleratedmobilepageurl/v1.ts
```diff
@@ -16,7 +16,9 @@
 
 /* jshint maxlen: false */
 
-import {createAPIRequest} from '../../lib/apirequest';
+import {
```
JustinBeckwith: Why did it get formatted like that?  Guessing clang will just put it back after we run `gts fix` against it.

jmdobry: want me to run `gts fix` on this PR?

JustinBeckwith: If you can - guessing that will be very rough

jmdobry: if I have to do it 1 file at a time...yeah...no. Looks like the templates just need to be fixed.

Issue #873, opened 2017-11-25
Replace swig with a maintained module


JustinBeckwith:
Swig is abandoned, and hasn't had a release in over 3 years:
https://github.com/paularmstrong/swig

There are a few transitive dependencies that need to be updated as well.

JustinBeckwith:
Fixed in #874

PR #874, opened 2017-11-25
chore: move from swig to nunjucks for templates

JustinBeckwith:
Addresses #873.  Swig is unmaintained, and has vulnerabilities with on it's transitive dependencies.  This moves us to [nunjucks](https://mozilla.github.io/nunjucks/), which was the closest thing to Swig that's actively maintained.  Also took this as an opportunity to move from js_beautify over to clang-format to be consistent with our toolchain.   

Most of the interesting changes that need scrutiny are at the bottom.


scripts/generator.ts
```diff
@@ -74,8 +48,8 @@ export class Generator {
    * @param  {string} str String to process
    * @return {string}     Single line string processed
    */
-  private oneLine(str: string) {
-    return str.replace(/\n/g, ' ');
+  private oneLine(str?: string) {
+    return str ? str.replace(/\n/g, ' '): '';
```
kjin: Nit: space before `:`

scripts/generator.ts
```diff
@@ -99,9 +73,11 @@ export class Generator {
    */
   private getAPIs(items) {
```
kjin: While you're editing this code, can you add a type for `items`?

scripts/generate.ts
```diff
@@ -31,28 +31,28 @@ const argv = minimist(process.argv.slice(2));
...
-    gen.generateAllAPIs(e => {
-      if (e) {
-        throw e;
-      }
-      console.log('Finished generating APIs!');
-    });
-  });
-}
+gen.generateIndex(err => console.error);
+// if (args.length) {
```
kjin: Should this commented code be removed?

scripts/generator.ts
```diff
@@ -267,13 +244,15 @@ export class Generator {
       }
       apis[file] = {};
       fs.readdirSync(path.join(apisPath, file)).forEach(version => {
-        apis[file][version] = path.parse(version).name;
+        const parts = path.parse(version);
+        if (!version.endsWith('.d.ts') && parts.ext === '.ts') {
+          apis[file][version] = path.parse(version).name;
```
kjin: Use `parts.name` here

scripts/generator.ts
```diff
@@ -267,13 +244,15 @@ export class Generator {
...
+        }
       });
     });
-
-    const result = swig.render(indexTemplateContents, {locals: {apis}});
-    const contents = js_beautify(result, BEAUTIFY_OPTIONS);
-    fs.writeFile(indexPath, contents, {encoding: 'utf8'}, err => {
+    console.log(apis);
```
kjin: Remove this log line?

scripts/generator.ts
```diff
@@ -267,13 +244,15 @@ export class Generator {
...
+    const result = this.env.render('index.njk', {apis: apis});
```
kjin: Nit: `{apis}` should work here

scripts/generate.ts
JustinBeckwith: yikes, thanks for catching

scripts/generator.ts
JustinBeckwith: Ran `gts fix` on the file

scripts/generator.ts
JustinBeckwith: Method wasn't even being used lol, deleted

scripts/generator.ts
JustinBeckwith: good catch

scripts/generator.ts
JustinBeckwith: removed

scripts/generator.ts
JustinBeckwith: fixed with the `gts fix`

scripts/generator.ts
```diff
@@ -34,38 +33,24 @@ const DISCOVERY_URL = argv['discovery-url'] ?
 const FRAGMENT_URL =
     'https://storage.googleapis.com/apisnippets-staging/public/';
 
-const API_TEMPLATE = './templates/api-endpoint.ts';
-const INDEX_TEMPLATE = './templates/index.ts';
-const BEAUTIFY_OPTIONS = {
```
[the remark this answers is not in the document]
JustinBeckwith: The previous code was reading in the contents of the template, and doing a render based on a string representation.  The new code just needs the file path to the template, and handles reading the file from disk for you :)  The beautify options went away because I got rid of beautify, and replaced it with clang-format.

JustinBeckwith:
The generated TypeScript does get compiled, and we have tests that cover some portion of the generated libraries.  The drive tests specifically cover a lot of ground here.  From spot checking, it does look like almost all of the changes are cosmetic.

PR #870, opened 2017-11-24
chore: run gts check on posttest

JustinBeckwith:

PR #877, opened 2017-11-28
chore: update generated APIs to pass format/lint

JustinBeckwith:
Cosmetic changes to get generated APIs passing our lint rules.  PR in #870 will add test to validate it.


templates/api-endpoint.njk
```diff
@@ -59,19 +59,19 @@ function {{ Name }} (options) {
 {% if api.schemas %}
 {% for schemaName, schema in api.schemas|dictsort %}
 /**
- * @typedef {{ schema.id }}
+ * @typedef {{ schema.id|oneLine }}
```
ofrobots: teeny nit: everywhere else you use spaces around the `|`. Use the same style.

PR #878, opened 2017-11-29
chore: exclude tests from codecov

JustinBeckwith:


JustinBeckwith:
It seemed generally like including tests in the coverage was generating noise.  It also prevents us from getting an accurate reading of our coverage, as we aim to increase it on the library.  

By default, test files are excluded from nyc:
https://github.com/istanbuljs/nyc#excluding-files

The only reason ours were included is that we're using TypeScript.  If others feel strongly, happy to close this out.

ofrobots:
Already LGTM. Feel free to land given that the rationale is documented now.

package.json
```diff
@@ -1,7 +1,8 @@
 {
   "nyc": {
     "exclude": [
-      "apis"
+      "apis",
+      "test"
```
jinwoo: Sorry for the late comment, but I think this should be `"build/test"` to property excluded, isn't it?

PR #883, opened 2017-12-02
chore: code reorganization

JustinBeckwith:
This change moves our source files under `src`, our build files under `build`, and uses the base `tsconfig-google.json` config.  It makes this library more consistent with our other libraries, and makes it easier to work with generated files.


JustinBeckwith:
@alexander-fenster Would be great if you could start reviewing these :)

src/scripts/generator.ts
```diff
@@ -222,7 +223,7 @@ export class Generator {
 
   generateIndex(callback: Function) {
     const apis = {};
-    const apisPath = path.join(__dirname, '../apis');
+    const apisPath = path.join(__dirname, '../../../src/apis');
```
alexander-fenster: move src/apis to constant as well?

JustinBeckwith: fixed!

white0ut:
Please note that this change broke link refs for KMS: https://cloud.google.com/kms/docs/reference/libraries#client-libraries-install-nodejs

You may want to check with other products referencing here to make sure they aren't broken.

JustinBeckwith:
Yar, thanks for the heads up.  @jmdobry is there an easy way to search the docs for links to this repo?

PR #891, opened 2017-12-11
Upgrade to the latest google-auth-library

JustinBeckwith:
The latest version of the google-auth-library makes a few breaking changes. 
- [x] Get the tests passing
- [x] Code cleanup
- [x] Update the documentation
- [x] Switch to the real release

Note - this is ready!  Please take another look if interested :)

### DIFF
--- a/src/lib/apirequest.ts
+++ b/src/lib/apirequest.ts
@@ -11,9 +11,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {DefaultTransporter} from 'google-auth-library/lib/transporters';
+import {AxiosRequestConfig} from 'axios';
+import {DefaultTransporter} from 'google-auth-library';
+import * as qs from 'qs';
 import * as stream from 'stream';
 import * as parseString from 'string-template';
+import * as uuid from 'uuid';
 
 // tslint:disable-next-line: no-any
 function isReadableStream(obj: any) {
@@ -51,10 +54,10 @@ function getMissingParams(params, required) {
  * @param  {Function} callback   Callback when request finished or error found
  * @return {Request}             Returns Request object or null
  */
-export function createAPIRequest(parameters, callback) {
-  let req, body, missingParams;
+export function createAPIRequest(parameters, callback): void {
+  let missingParams;
   let params = parameters.params;
-  let options = Object.assign({}, parameters.options);
+  let options: AxiosRequestConfig = Object.assign({}, parameters.options);
 
   // If the params are not present, and callback was passed instead,
   // use params as the callback and create empty params.
@@ -120,6 +123,15 @@ export function createAPIRequest(parameters, callback) {
     parameters.mediaUrl = parseString(parameters.mediaUrl, params);
   }
 
+  // When forming the querystring, override the serializer so that array
+  // values are serialized like this:
+  // myParams: ['one', 'two'] ---> 'myParams=one&myParams=two'
+  // This serializer also encodes spaces in the querystring as `%20`,
+  // whereas the default serializer in axios encodes to a `+`.
+  options.paramsSerializer = (params) => {
+    return qs.stringify(params, {arrayFormat: 'repeat'});
+  };
+
   // delete path parameters from the params object so they do not end up in
   // query
   parameters.pathParams.forEach(param => {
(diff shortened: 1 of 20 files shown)


JustinBeckwith:
cc @alexander-fenster

test/test.options.ts
```diff
@@ -44,108 +51,162 @@ describe('Options', () => {
...
+      if (err) return done(err);
       // If the default param handling is broken, query might be undefined, thus
       // concealing the assertion message with some generic "cannot call
       // .indexOf of undefined"
-      query = req.uri.query || '';
+      let query = Utils.getQs(res) || '';
       assert.notEqual(
           query.indexOf('myParam=123'), -1, 'Default param not found in query');
-      done();
+      nock.enableNetConnect();
```
JustinBeckwith: @ofrobots for some reason, I can't seem to get enableNetConnect() to work here.  I keep getting a `Error: Nock: No match for request`, which AFAICT shouldn't happen.  I could use a second look from someone here.

alexander-fenster:
LGTM but please take a look at the CI error

src/lib/apirequest.ts
```diff
@@ -136,48 +147,49 @@ export function createAPIRequest(parameters, callback) {
...
+      // options.multipart = [
+      //   {
+      //     'Content-Type': 'application/json',
+      //     body: JSON.stringify(resource)
+      //   }, {
+      //     'Content-Type': media.mimeType || (resource && resource.mimeType)
+      //     || defaultMime, body: media.body  // can be a readable stream or
+      //     raw string!
+      //   }
+      // ];
```
ofrobots: nit: remove commented code.

src/lib/apirequest.ts
```diff
@@ -136,48 +147,49 @@ export function createAPIRequest(parameters, callback) {
...
       } else {
-        options.body = media.body;
+        options.data = media.body;
       }
     }
   } else {
-    options.json = resource ||
-        ((options.method === 'GET' || options.method === 'DELETE') ? true : {});
+    options.data = resource || undefined;
+    //((options.method === 'GET' || options.method === 'DELETE') ? true :
```
ofrobots: nit: remove commented code

src/lib/apirequest.ts
```diff
@@ -135,49 +147,62 @@ export function createAPIRequest(parameters, callback) {
...
           'Content-Type':
               media.mimeType || (resource && resource.mimeType) || defaultMime,
           body: media.body  // can be a readable stream or raw string!
         }
       ];
+      const boundary = uuid.v4();
+      const finale = `--${boundary}--`;
+      const rStream = new stream.PassThrough();
+      const isStream = isReadableStream(multipart[1].body);
+      headers['Content-Type'] = `multipart/related; boundary=${boundary}`;
```
ofrobots: It seems like we need a standalone module that provides multipart support on top of axios. [form-data](https://www.npmjs.com/package/form-data) might be relevant.

src/lib/discovery.ts
```diff
@@ -171,67 +180,68 @@ Discovery.prototype.log = function() {
...
-    if (err) {
-      return handleError(err, callback);
-    }
-
-    async.parallel(
-        resp.items.map(api => {
-          return (cb) => {
-            self.discoverAPI(api.discoveryRestUrl, (e, newApi) => {
+  transporter.request<DiscoverAPIsResponse>(
+      {url: discoveryUrl, headers}, (err, resp) => {
```
ofrobots: nit: is it possible to use the promise version here?

src/lib/discovery.ts
```diff
@@ -171,67 +180,68 @@ Discovery.prototype.log = function() {
...
+        if (err) {
+          return handleError(err, callback);
+        }
+        const items = resp.data.items;
+        async.parallel(
```
ofrobots: nit: is it possible to `Promise.all` on an array of promises; dropping the dependency on async? It is fine to tackle this in a follow-on though.

test/test.media.ts
```diff
@@ -3,7 +3,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//    http://www.apache.org/licenses/LICENSE-2.0
+//    http://www.apache.o rg/licenses/LICENSE-2.0
```
ofrobots: unintended change

test/test.path.ts
```diff
@@ -113,40 +115,100 @@ describe('Path params', () => {
...
-  it('should keep query params as is', () => {
-    let req =
-        localDrive.files.get({fileId: '123abc', hello: 'world'}, Utils.noop);
-    assert.equal(req.uri.query, 'hello=world');
-    req = remoteDrive.files.get({fileId: '123abc', hello: 'world'}, Utils.noop);
-    assert.equal(req.uri.query, 'hello=world');
+  it('should keep query params as is', (done) => {
+    const p = '/drive/v2/files/123abc?hello=world';
+    nock(Utils.baseUrl).get(p).reply(200);
+    localDrive.files.get({fileId: '123abc', hello: 'world'}, (err, res) => {
```
ofrobots: any reason not to use `pify` here like you do elsewhere?

nit: it would have been better if the pify changes were not included in this PR. It bundles optional changes with necessary changes making the patch more work to review.

test/test.transporters.ts
```diff
@@ -113,29 +118,27 @@ describe('Transporters', () => {
...
-    testHeaders(remoteDrive);
+  it('should add headers to the request from params', async () => {
+    return Promise.all([testHeaders(localDrive), testHeaders(remoteDrive)]);
   });
 
-  it('should automatically add content-type for POST requests', () => {
-    testContentType(localDrive);
-    testContentType(remoteDrive);
+  it('should automatically add content-type for POST requests', async () => {
+    return Promise.all(
```
ofrobots: question: is `Promise.all` necessary here? You could have awaited each of the test functions. Is it essential that these two tests run in parallel? This applies here and elsewhere the same pattern is used.

nit: don't use `async` unless the function needs to await.

src/lib/discovery.ts
JustinBeckwith: fixed!

src/lib/discovery.ts
JustinBeckwith: Async is used all over this mess.  Upgrading to promises is going to be a challenge in and of itself, so I'd rather leave this for a follow up.

test/test.media.ts
JustinBeckwith: weird.  nice catch!

test/test.path.ts
JustinBeckwith: I only updated the tests where I would have had to add substantial code otherwise.  In many places, we didn't need to follow through on the full request, because we were just checking properties on the returned req object.  Now that it doesn't return a request (rather a promise), it was A LOT of adding nock and pify to keep callback depth sane.

test/test.transporters.ts
JustinBeckwith: I updated them to use await

src/lib/apirequest.ts
JustinBeckwith: I filed an issue against the axios repo:
https://github.com/axios/axios/issues/1247

Based on their feedback, I'll either submit a PR or go build a standalone module.

ofrobots: possibly relevant: https://github.com/axios/axios/issues/789

test/test.path.ts
ofrobots: ack.

src/lib/discovery.ts
ofrobots: ack.

PR #916, opened 2018-01-10
fix: cleanup and fix samples

JustinBeckwith:
When trying to nail down #891, I noticed that most of the samples were in disrepair.  To be able to effectively test the new auth lib, we should probably land this first.


samples/blogger/blogger.js
```diff
@@ -13,18 +13,16 @@
 
 'use strict';
 
-var google = require('../../lib/googleapis');
-var blogger = google.blogger('v3');
-var util = require('util');
+const google = require('googleapis');
+const blogger = google.blogger('v3');
+const config = require('../config.json');
```
alexander-fenster: Normally we read api key from the environment variable, which makes it easier to test the samples code in the continuous integration (which we don't do now anyway). If it's too much change, I'm ok to leave it in the file.

samples/defaultauth.js
```diff
@@ -24,19 +24,20 @@ var compute = google.compute('v1');
  * method rather than creating your own JWT or Compute client directly.
  *
  * Note: In order to run on a local developer machine, it is necessary to download a private key
- * file to your machine, and to set a local environment variable pointing to the location of the
+ * file to your machine, and to set a local environment constiable pointing to the location of the
```
alexander-fenster: I'm not a native English speaker but it looks like replacing all `var` with `const` caused some unexpected consequences!

JustinBeckwith: *laughs* oops.  Nice catch.

samples/blogger/blogger.js
JustinBeckwith: It's not a huge deal.  Typically if I'm going to do that, I like to use `nconf` so I can support both file based and env based methods.  I'll take a shot at that tonight.

alexander-fenster:
Unrelated question: do we need to protect the master branch for this repo to require at least the Travis CI to pass (not mentioning circle)?

JustinBeckwith:
YES.  We should absolutely enable master branch protection.

alexander-fenster:
I have enabled branch protection for branch master in this repo.

PR #924, opened 2018-01-13
add github-specific files to .gitignore

MikailBag:

- [ ] `npm test` succeeds (must))


MikailBag:
As I can see, travis build fails on the same configurations as repository head.

MikailBag:
I signed it!

JustinBeckwith:
Thanks for the PR!  We recently decided to drop the `.npmignore` approach (#966), and instead use the opt-in model (#955).  With that change, I think we're pretty set here.  Thanks for letting us know this was a problem!

PR #926, opened 2018-01-15
chore: asyncify generator

JustinBeckwith:
The callback code in the generator was hard to reason about.  This change moves us towards async/await style code, and speeds things up quite a bit.


JustinBeckwith:
The weird casts to any will be fixed when this gets merged & released:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/22909

PR #932, opened 2018-01-17
release 25.0.0

alexander-fenster:
Releasing v25.0.0.


JustinBeckwith:
Right now something is up with Travis, but even before that - CI seemed to be broken for Node.js 4 and 6.  Usually this means there was an update in some sort of transitive dependency, and not having a package-lock.json (like we do in 8 and 9) has bitten us.  I haven't had time to debug the issue, but I think that's something we should do before cutting a release.

alexander-fenster:
I'll take a look. Also, do you mind changing travis to circle for this repo? I can do that.

JustinBeckwith:
I wouldn't mind at all.  Thank you :) That and setting up automatic releases were on the list.

alexander-fenster:
OK :)

PR #949, opened 2018-01-23
fix test

alexander-fenster:
Fixes failing test for node4 and node6. 

- [x] `npm test` succeeds

### DIFF
--- a/test/test.auth.ts
+++ b/test/test.auth.ts
@@ -210,6 +210,12 @@ describe('OAuth2 client', () => {
     };
 
     nock(Utils.baseUrl).get('/drive/v2/files/wat').reply(200);
+    nock(Utils.baseUrl).post('/oauth2/v4/token').reply(200, {
+      'access_token': 'abc123',
+      'refresh_token': 'abc',
+      'expiry_date': tenSecondsFromNow,
+      'token_type': 'Bearer'
+    });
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
That's weird 😂

JustinBeckwith:
Line 198 of test.auth.ts, we create nock for the old oauth token endpoint.  Please just replace that code with the new code you've added :)

alexander-fenster:
done, PTAL

test/test.auth.ts
```diff
@@ -210,6 +206,12 @@ describe('OAuth2 client', () => {
     };
 
     nock(Utils.baseUrl).get('/drive/v2/files/wat').reply(200);
+    let scope = nock(Utils.baseUrl).post('/oauth2/v4/token').reply(200, {
```
JustinBeckwith: Sorry for missing this earlier.  You can just use `.times(s)` and a `const` here to ensure it runs twice :)

alexander-fenster: After some looking at the code, I guess the point is that it should **not** run at all. That's what happens in node8 and node9, and what does not happen in node4 and node6.

JustinBeckwith:
Going to go ahead and close this one out since we're back up.

PR #954, opened 2018-01-24
Create .appveyor.yml

alexander-fenster:
Let's use this one as well (but not requiring it to pass because sometimes it takes too long).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,20 @@
+environment:
+  matrix:
+  - nodejs_version: 8
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - npm install -g npm # Force using the latest npm to get dedupe during install
+  - set PATH=%APPDATA%\npm;%PATH%
+  - npm install --force --ignore-scripts
+
+test_script:
+  - node --version
+  - npm --version
+  - npm rebuild
+  - npm test
+
+build: off
+
+matrix:
+  fast_finish: true
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,9 +4,8 @@ environment:
 
 install:
   - ps: Install-Product node $env:nodejs_version
-  - npm install -g npm # Force using the latest npm to get dedupe during install
   - set PATH=%APPDATA%\npm;%PATH%
-  - npm install --force --ignore-scripts
+  - npm install
 
 test_script:
   - node --version


.appveyor.yml
```diff
@@ -0,0 +1,20 @@
+environment:
+  matrix:
+  - nodejs_version: 8
```
JustinBeckwith: Why just 8, and not 4/6/8/9?

.appveyor.yml
```diff
@@ -0,0 +1,20 @@
...
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - npm install -g npm # Force using the latest npm to get dedupe during install
```
JustinBeckwith: Why? I'd assume we want to use the version of npm that comes with the node release.

.appveyor.yml
```diff
@@ -0,0 +1,20 @@
...
+  - set PATH=%APPDATA%\npm;%PATH%
+  - npm install --force --ignore-scripts
```
JustinBeckwith: Sorry for all the questions.  Why --force and --ignore-scripts?

alexander-fenster: Just copied the one we use in googleapis/nodejs-* projects. To answer your questions.
(a) as the comment states - `Force using the latest npm to get dedupe during install`. Not sure if it's actual, I'm OK to remove.
(b) I'm pretty sure some scripts work only on Linux and we don't want to fail if any of those fail.

JustinBeckwith: Wouldn't those scripts fail an `npm install` on Windows though?

.appveyor.yml
alexander-fenster: It runs 2 builds per branch commit, sequentially, and is slower than Circle - we cann add 4/6/9 but it will take hours to finish.

.appveyor.yml
alexander-fenster: as the comment states - `Force using the latest npm to get dedupe during install`. Not sure if it's actual, I'm OK to remove.

.appveyor.yml
alexander-fenster: Not sure about `--force`: as I said, I just copied it from another place :) We can try removing it and see what happens.

.appveyor.yml
JustinBeckwith: Wow lol.  Ok, just 8 it is :)

.appveyor.yml
JustinBeckwith: Lets start with out it and add it back if needed

.appveyor.yml
JustinBeckwith: Lets start without them, and add them back if needed.  Lets keep it as close as possible to the actual developer workflow

alexander-fenster:
OK, let's see if it works.

PR #962, opened 2018-01-25
try some appveyor magic

alexander-fenster:
Windows builds fail being unable to run `npm` using `childprocess.spawn`. Details [here](https://github.com/nodejs/node/issues/3675). As a quick solution, passing `{ shell: true }` to `spawn` helps (I guess it's OK for the test code).


alexander-fenster:
I feel extremely proud of setting up Appveyor on that same day when we first had a Windows incompatibility.

JustinBeckwith:
Nice catch!  Sorry for leaving this busted in the previous PR.

PR #970, opened 2018-01-26
chore: make a few APIs async

JustinBeckwith:
This change does a few things:
- Eliminates the `async` and `power-assert` npm depdendencies
- Introduces a few async APIs in the generator
- Basic cleanup

One step closer to noImplicitAny.

### DIFF
--- a/test/test.media.ts
+++ b/test/test.media.ts
@@ -16,9 +16,7 @@ import * as fs from 'fs';
 import * as nock from 'nock';
 import * as path from 'path';
 import * as pify from 'pify';
-
 import {GoogleApis} from '../src';
-
 import {Utils} from './utils';
 
 const boundaryPrefix = 'multipart/related; boundary=';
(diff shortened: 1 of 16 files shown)


test/test.media.ts
```diff
@@ -72,28 +71,13 @@ describe('Media', () => {
...
-        (err, apis) => {
-          if (err) {
-            return done(err);
-          }
-          remoteDrive = apis[0];
-          remoteGmail = apis[1];
-          nock.disableNetConnect();
-          done();
-        });
+    remoteDrive = await Utils.loadApi(google, 'drive', 'v2');
```
alexander-fenster: I'm just excited how async/await makes the code so much easier.

alexander-fenster: (but probably a little bit slower as we're not parallel here anymore)

test/utils.ts
```diff
@@ -26,14 +27,9 @@ export abstract class Utils {
...
-      google: GoogleApis, name: string, version: string, options, cb) {
-    if (typeof options === 'function') {
-      cb = options;
-      options = {};
-    }
-    return google.discoverAPI(
-        Utils.getDiscoveryUrl(name, version), options, cb);
+  static async loadApi(
+      google: GoogleApis, name: string, version: string, options = {}) {
+    return google.discoverAPI(Utils.getDiscoveryUrl(name, version), options);
```
alexander-fenster: should we await for the result here, then return?

JustinBeckwith: nah, don't need to.  I actually deleted the `async` from the fn, because it just returns a promise (which is awaitable) directly.

test/test.media.ts
JustinBeckwith: Thanks!  The serial thing was an accident.  Submitting a fix for that now :)

PR #1010, opened 2018-02-20
fix: move pify to deps

### LINKED ISSUE #1007: 'pify' should not be listed in 'devDependencies' list
https://github.com/google/google-api-nodejs-client/blob/a46f7302ce4c028fb4c5615a0a58432a87411fc6/package.json#L132

When installing packages in production environment (`NODE_ENV=production yarn install`), app breaks in runtime, not finding **pify** module to load.
I guess this should be fixed.

JustinBeckwith:
Resolves #1007

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "googleapis",
-  "version": "25.0.0",
+  "version": "26.0.1",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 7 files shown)


jinwoo:
I'm not sure what's up with the test failure in node4.

JustinBeckwith:
This turned out to be way more of an adventure than I was expecting, but here we are.

PR #1016, opened 2018-02-22
fix: raise default of maxContentLength

### LINKED ISSUE #991: Unable to upload more than 10MB
When using the example code for uploading videos to youtube using the api (https://developers.google.com/youtube/v3/docs/videos/insert) i'm getting the following error for all videos over 10MB:

```
The API returned an error:  Error: Request body larger than maxBodyLength limit
    at RedirectableRequest.write (C:\Users\Malte\Googl…w-redirects\ind…:66)
    at PassThrough.ondata (_stream_readable.js:642)
    at emitOne (events.js:125)
    at PassThrough.emit (events.js:221)
    at addChunk (_stream_readable.js:265)
    at readableAddChunk (_stream_readable.js:252)
    at PassThrough.Readable.push (_stream_readable.js:209)
    at PassThrough.Transform.push (_stream_transform.js:146)
    at PassThrough.afterTransform (_stream_transform.js:87)
    at PassThrough._transform (_stream_passthrough.js:42)
```

~I was able to fix this by editing https://github.com/olalonde/follow-redirects/blob/master/index.js#L226 of the follow-redirects module to something bigger as this is stopping all bigger requestbodys and therefor stop the uploading of most videos.~

Edit: A better way to fix this is to install Version 1.2.6 of follow-redirect as this is the newest version without this new "feature".

See also this commit that implemented this restriction to follow-redirects https://github.com/olalonde/follow-redirects/commit/847a47fb00ea2dcb7a8886ca16d9d82dc6accb5d

JustinBeckwith:
Resolves #991

### DIFF
--- a/src/lib/apirequest.ts
+++ b/src/lib/apirequest.ts
@@ -21,6 +21,8 @@ import * as uuid from 'uuid';
 import {APIRequestParams} from './api';
 import {SchemaParameters} from './schema';
 
+const maxContentLength = Math.pow(2, 31);
+
 function isReadableStream(obj: stream.Readable|string) {
   return obj instanceof stream.Readable && typeof obj._read === 'function';
 }
@@ -193,7 +195,7 @@ export function createAPIRequest<T>(
   // We need to set a default content size, or the max defaults
   // to 10MB.  Setting to 2GB by default.
   // https://github.com/google/google-api-nodejs-client/issues/991
-  options.maxContentLength = options.maxContentLength || Math.pow(2, 31);
+  options.maxContentLength = options.maxContentLength || maxContentLength;
 
   // Combine the AxiosRequestConfig options passed with this specific
   // API call witht the global options configured at the API Context
--- a/src/lib/apirequest.ts
+++ b/src/lib/apirequest.ts
@@ -190,6 +190,10 @@ export function createAPIRequest<T>(
 
   options.headers = headers;
   options.params = params;
+  // We need to set a default content size, or the max defaults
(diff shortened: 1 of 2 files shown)


src/lib/apirequest.ts
```diff
@@ -190,6 +190,10 @@ export function createAPIRequest<T>(
 
   options.headers = headers;
   options.params = params;
+  // We need to set a default content size, or the max defaults
+  // to 10MB.  Setting to 2GB by default.
+  // https://github.com/google/google-api-nodejs-client/issues/991
+  options.maxContentLength = options.maxContentLength || Math.pow(2, 31);
```
jinwoo: Is it necessary to increase the default max? In my opinion that is the decision to be made by the application. Isn't it good enough to just allow the app developers to configure it, which is already supported?

I don't know but 2GB seems too high to me as the default value.

JustinBeckwith: The choice of 10MB was a somewhat random decision on the part of a transitive dependency we use.  This was not a conscious choice by us, or Axios. The change is very much semver major, and frankly - not something I think is necessary.  Moving to 2GB was my way of saying 'nah, this is fine'.  For the rare developer that wants to upload more than that... they have the option :P

jinwoo: OK. I don't feel strongly about the default max anyway. Just one nit. Can you define a constant instead of inlining `Math.pow(2, 31)` here?

JustinBeckwith:
Absolutely!

Issue #999, opened 2018-02-13
sheets quickstart.js needs updating to reflect the es6 'optimizations' 


smurftruesmart:
Hi,

quickstart.js ( taken from https://developers.google.com/sheets/api/quickstart/nodejs ) currently fails when trying to reference google.sheets('v4') due to changes in google-api-nodejs-client 26.0.0.

    var google = require('googleapis');

should be replaced with:

    const {google} = require('googleapis');

otherwise the following error occurs:

/var/nodejs/sheets/quickstart.js:105
  var sheets = google.sheets('v4');
                      ^

TypeError: google.sheets is not a function
    at listMajors (/var/nodejs/sheets/quickstart.js:105:23)
    at /var/nodejs/sheets/quickstart.js:78:7
    at /var/nodejs/sheets/node_modules/google-auth-library/lib/auth/oauth2client.js:95:13
    at Request._callback (/var/nodejs/sheets/node_modules/google-auth-library/lib/transporters.js:113:17)
    at Request.self.callback (/var/nodejs/sheets/node_modules/request/request.js:186:22)
    at Request.emit (events.js:160:13)
    at Request.<anonymous> (/var/nodejs/sheets/node_modules/request/request.js:1163:10)
    at Request.emit (events.js:160:13)
    at IncomingMessage.<anonymous> (/var/nodejs/sheets/node_modules/request/request.js:1085:12)
    at Object.onceWrapper (events.js:255:19)

Hope this helps.

Thanks!

cepm-nate:
Same thing for the google.calendar('v3') in the quickstart.

JustinBeckwith:
Apologies for the issues folks.  We added a sample in #1022, and we're waiting on a docs change to roll out with the update.  For other Googlers, this is b/73865350.

grant:
The quickstart on https://developers.google.com/sheets/api/quickstart/nodejs should now be runnable using `@25.*` (instructions are updated).
Updating to `@26`+ is WIP.

JustinBeckwith:
👋 Hey @grant!  I wrote a sample for this in the repo here:
https://github.com/google/google-api-nodejs-client/blob/master/samples/sheets/quickstart.js

I'm trying to move the samples on codesite over to GitHub so we can write tests against them, preventing this kind of thing from happening again :)  I also made a few suggestions over in b/73865350 that should be added as well. Thanks!

grant:
[...]

PR #1022, opened 2018-02-24
docs: add the sheets quickstart

JustinBeckwith:
We're seeing a bunch of reports of old docs from codesite.  This is the first step to fixing #999.   We will need to follow up with tests, and then submit a CL to embed the code snippet.

### DIFF
--- a/samples/sheets/quickstart.js
+++ b/samples/sheets/quickstart.js
@@ -1,0 +1,89 @@
+// Copyright 2018, Google, LLC.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// [START main_body]
+const {google} = require('googleapis');
+const express = require('express');
+const opn = require('opn');
+const path = require('path');
+const fs = require('fs');
+
+const keyfile = path.join(__dirname, 'credentials.json');
+const keys = JSON.parse(fs.readFileSync(keyfile));
+const scopes = ['https://www.googleapis.com/auth/spreadsheets.readonly'];
+
+// Create an oAuth2 client to authorize the API call
+const client = new google.auth.OAuth2(
+  keys.web.client_id,
+  keys.web.client_secret,
+  keys.web.redirect_uris[0]
+);
+
+// Generate the url that will be used for authorization
+this.authorizeUrl = client.generateAuthUrl({
+  access_type: 'offline',
+  scope: scopes
+});
+
+// Open an http server to accept the oauth callback. In this
+// simple example, the only request to our webserver is to
+// /oauth2callback?code=<code>
+const app = express();
+app.get('/oauth2callback', (req, res) => {
+  const code = req.query.code;
+  client.getToken(code, (err, tokens) => {
+    if (err) {
+      console.error('Error getting oAuth tokens:');
+      throw err;
+    }
+    client.credentials = tokens;
+    res.send('Authentication successful! Please return to the console.');
+    server.close();
+    listMajors(client);
+  });
+});
+const server = app.listen(3000, () => {
+  // open the browser to the authorize url to start the workflow
+  opn(this.authorizeUrl, { wait: false });
+});
+
+/**
+ * Print the names and majors of students in a sample spreadsheet:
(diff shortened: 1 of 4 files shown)


samples/sheets/quickstart.js
```diff
@@ -0,0 +1,89 @@
...
+    auth: auth,
+    spreadsheetId: '1BxiMVs0XRA5nFMdKvBdBZjgmUUqptlbs74OgvE2upms',
+    range: 'Class Data!A2:E',
+  }, (err, res) => {
+    if (err) {
+      console.error('The API returned an error.');
+      throw err;
+    }
+    const rows = res.data.values;
+    if (rows.length == 0) {
```
alexander-fenster: ===

alexander-fenster: Looks like the CI is complaining about the same line :)

grant:
@JustinBeckwith @alexander-fenster 

Thanks for the Sheets addition. I want to let you know that I'm releasing G Suite Node samples (quickstarts + snippets) for ~6 G Suite products in a week or so. I appreciate the effort to updating this one and putting it on GitHub, but it looks like this code is directly pulled from the quickstart with some touchups.

I've since added the dependency version to the [sheets quickstart](https://developers.google.com/sheets/api/quickstart/nodejs) so that it does break. It will be updated and linked to GitHub when the **`gsuitedevs/node-samples`** repo is published.

In order to not have duplicate code, and to match the style of the `samples/` that doesn't have quickstarts like this, would you consider adding a cross link to the G Suite samples instead of having 2 versions when the node-samples repo is published? It seems like adding this quickstart is trying to patch the root problem.

JustinBeckwith:
>  I want to let you know that I'm releasing G Suite Node samples (quickstarts + snippets) for ~6 G Suite products in a week or so.
Yes!  This is amazing!  

> it looks like this code is directly pulled from the quickstart with some touchups.
That's because I submitted the CL to update the quickstart first 🤣 I wanted to get that updated quickly (people were getting led astray) while we worked on getting real samples checked in and tested.  

> I've since added the dependency version to the sheets quickstart so that it does break.
The other approach we could take here is to instruct users to install at head, but keep all the samples in this repo so they're always tested.  I'm a humble PM volunteering time here, and the tests in this lib far from cover all scenarios.  I want to make sure our samples are tested here so I know that I'm not breaking anyone.  

> In order to not have duplicate code, and to match the style of the samples/ that doesn't have quickstarts like this, would you consider adding a cross link to the G Suite samples instead of having 2 versions when the node-samples repo is published? It seems like adding this quickstart is trying to patch the root problem.
I would prefer to move all our node-samples for this library over here, for the reasons stated above.  As we do make breaking changes, we can make sure we're fixing the docs in the same PR that introduces the problem.  Also, it's always nice to have more node.js folks to look at PRs in these repos :)

This appears to have a lot of moving parts.  If you'd like to chat on Monday, happy to hop on a GVC!

grant:
Thanks for volunteering time. I'm sure some sample overlap is fine.

> That's because I submitted the CL to update the quickstart first 🤣 I wanted to get that updated quickly (people were getting led astray) while we worked on getting real samples checked in and tested.

The best way to update the quickstart would be to send a message to the DevSite author saying that the docs sample is broken. Copying this one sample into another repo is not a great fix. There were 5 other product samples that I updated as well that weren't reported. Installing at HEAD is not a reliable solution for every language/product sample.

---

Some more background:

The G Suite samples are planned to be in a similar structure to GCP's samples:

https://github.com/GoogleCloudPlatform/nodejs-docs-samples

There are ~10 languages, so any solution of adding samples to client repos would have to be consistent will all clients:
https://github.com/google?utf8=%E2%9C%93&q=google-api-&type=&language=

There are complications (like no python client, multiple sample repos) that makes adding samples to client repos difficult. It's really a more complicated discussion that was talked about with OSPO/github-requests@ a few months ago.

If we want to start a discussion about consolidating all Google Node samples in this repo, including [GCP](https://github.com/GoogleCloudPlatform/nodejs-docs-samples), [this client library](https://github.com/google/google-api-nodejs-client), G Suite, [googlemaps](https://github.com/googlemaps/js-samples), etc. then we should start a conversation with Max/OSPO and all sample repo maintainers.

The main point I'd like to get across is that copying one G Suite sample from DevSite so there are **two versions** is not a great solution to fixing this problem. Simply ask the Google author for the docs to be fixed.

JustinBeckwith:
I'm in total agreement that having two versions of these samples is a bad idea.  I didn't know y'all were writing and maintaining these samples - this is great!  

Now on the repo where the samples live - this is probably a discussion that needs to include @jonparrott @jmdobry and @ace-n.  Folks - what's the future of the `nodejs-docs-samples` repo?  I was under the understanding that we're actively trying to pull things out of there, and move samples closer to the repo with the actual implementations.  Is that accurate?

grant:
Note: All G Suite docs and samples are maintained by G Suite DevRel.

I'll set up a meeting next week for coordinating these samples. Be wary that G Suite and `@google-cloud` are a bit different, and that any solution proposed should cover .NET, Android, Apps Script, Go, iOS, Java, JS (client-side), Node, PHP, Python and Ruby equally.

Issue #1024, opened 2018-02-25
Downloading a file using the drive API node.js example doesn't work


matlsn:
Using the example found on [the drive REST api page](https://developers.google.com/drive/v3/web/manage-downloads) doesn't work as intended.

Node returns "cannot read property 'on' of undefined".

thomann061:
See here:
https://github.com/google/google-api-nodejs-client/issues/963

JustinBeckwith:
Greetings folks!  We're in the process of working out how we manage our samples.  Until that gets sorted out - I've put together a working set of samples here:
https://github.com/google/google-api-nodejs-client/pull/1028

Thanks for your patience!

JustinBeckwith:
Greetings folks!  We've merged in the samples with a working sample in this repo.  Sadly we use a different bug system for bugs on developers.google.com.  I went ahead and filed a bug here for you:
https://issuetracker.google.com/issues/74344655

Since this one is out of our hands, I'm going to close it out.  Let me know if you have any more problems!

PR #1028, opened 2018-02-28
test: refactor and test drive samples

JustinBeckwith:
Updates the drive samples, adds region tags for the docs, and actually tests the samples a bit.  First step towards #1024.

cc @jonparrott


JustinBeckwith:
cc @grant

PR #1029, opened 2018-03-01
chore: use make for builds

JustinBeckwith:
So this is an experiment :) I wanted to see what moving to Make would look like for our builds.  Some of the benefits:
- I think this is much easier to read
- We can have better control on which tasks get run when
- We can introduce things like timestamp based caching for formatting or builds

Some of the drawbacks:
- ~~~It's unclear if it works on windows or not~~~
- There is a bit of a learning curve (I am a novice at best)

Really just interested in other thoughts here :)


JustinBeckwith:
@google/google-node-team @ofrobots @jonparrott @jmdobry

jinwoo:
I don't think Windows has the `make` command by default. Something like Cygwin or MinGW should be installed.

And you can't use commands in `node_modules/.bin` from Makefile. `npm` modifies `$PATH` to make them available, but `make` doesn't know about that directory.

JustinBeckwith:
@ofrobots @google/google-node-team this is ready for another look.  This took the total test time from ~60s down to ~25s with the `-j` flag.  I'm pretty sure I can squeeze more out of it too :)

PR #1030, opened 2018-03-01
test: fixing possible cross-device rename causing EXDEV

alexander-fenster:
`test.kichen.ts` fails if `/tmp` is a different filesystem (which is quite common). The reason for that is that `fs.rename` is just a wrapper to system `rename(2)` which is unable to move a file between filesystems (returns EXDEV). Replacing `fs.rename` with copying followed by `fs.unlink`.

- [x] `npm test` succeeds


package-lock.json
```diff
@@ -705,17 +705,6 @@
         "underscore-contrib": "0.3.0"
       }
     },
-    "chai": {
```
JustinBeckwith: Why are there any package-lock.json changes with this PR?

test/test.kitchen.ts
```diff
@@ -20,13 +20,24 @@ import {ncp} from 'ncp';
 import * as pify from 'pify';
 import * as tmp from 'tmp';
 
-const rename = pify(fs.rename);
```
JustinBeckwith: This generally feels like a bug in node.js @ofrobots @mylesborins.  Would you mind filing a bug in their repo and linking to it here?

test/test.kitchen.ts
```diff
@@ -20,13 +20,24 @@ import {ncp} from 'ncp';
...
+const unlink = pify(fs.unlink);
 const ncpp = pify(ncp);
 const keep = !!process.env.GANC_KEEP_TEMPDIRS;
 const stagingDir = tmp.dirSync({keep, unsafeCleanup: true});
 const stagingPath = stagingDir.name;
 const pkg = require('../../package.json');
 
+function copyFile(src, dst) {
```
JustinBeckwith: `src` and `dst` need type annotations

JustinBeckwith: Is there a reason we can't use `fs.copyFile` here?
https://nodejs.org/api/fs.html#fs_fs_copyfile_src_dest_flags_callback

Either way, we need a comment explaining why we re-invented this method, along with that aforementioned link to the nodejs/node issue.

alexander-fenster: We cannot, fs.copyFile is added in 8.5.0. I actually tried using it but CI quickly explained that I was wrong :)

test/test.kitchen.ts
alexander-fenster: That's not a bug, rather a feature. Node.js documentation clearly states that `fs.rename` is 'Asynchronous `rename(2)`.' `rename(2)` [manual](http://man7.org/linux/man-pages/man2/rename.2.html) states it will return error: 
```
EXDEV  oldpath and newpath are not on the same mounted filesystem.
```

JustinBeckwith: Haha bull hockey :) The original intent may have been to bind themselves to unix, but that's an awful limitation on the `fs` API.  This is a usability bug :)

test/test.kitchen.ts
alexander-fenster: Added.

jinwoo:
FWIW I found a package for renaming across different devices: https://github.com/andrewrk/node-mv.

alexander-fenster:
I guess using `mv` module makes sense.

test/test.kitchen.ts
alexander-fenster: Removed this code - now using `mv` module.

package-lock.json
alexander-fenster: Now that I added a module I guess the `package-lock.json` changes are expected (before that, it was my mistake as a result of `git commit -a -m`)

alexander-fenster:
They already have one (closed as wontfix back in 2012): https://github.com/nodejs/node-v0.x-archive/issues/2703 - `rename()` is atomic and low-level and they will not make it do more than that.

test/test.kitchen.ts
ofrobots: I guess this suggests that `fs` is a bit too low level. It is not clear to me that Node.js needs to provide the low level API as well as a high level usable API. Node core *does* have a strong force of standarization – as there isn't any other higher level library that people by default reach out to (e.g. `lodash` but for `fs`). Perhaps we need to do a better job steering users towards higher level alternatives.

JustinBeckwith: I understand why it's implemented this way, I just happen to disagree with the decision :)  Good conversation happening over here: 
https://github.com/nodejs/node/issues/19077

PR #1045, opened 2018-03-07
chore: remove example keyfiles

JustinBeckwith:
These are making it kinda painful to work on the library.  I'd rather just entirely gitignore them, since users need to download them anyway.


alexander-fenster:
Why only node8 CI fails?

JustinBeckwith:
@google/google-node-team  @alexander-fenster ready for another look :)

PR #1065, opened 2018-03-13
feat: request gzip responses

### LINKED ISSUE #10: Support gzip

JustinBeckwith:
Resolves #10

### DIFF
--- a/src/lib/apirequest.ts
+++ b/src/lib/apirequest.ts
@@ -23,6 +23,10 @@ import {SchemaParameters} from './schema';
 
 const maxContentLength = Math.pow(2, 31);
 
+// tslint:disable-next-line no-var-requires
+const pkg = require('../../../package.json');
+const USER_AGENT = `google-api-nodejs-client/${pkg.version} (gzip)`;
+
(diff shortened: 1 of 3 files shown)


src/lib/apirequest.ts
```diff
@@ -205,6 +209,8 @@ export function createAPIRequest<T>(
   // to 10MB.  Setting to 2GB by default.
   // https://github.com/google/google-api-nodejs-client/issues/991
   options.maxContentLength = options.maxContentLength || maxContentLength;
+  options.headers['Accept-Encoding'] = 'gzip';
```
jinwoo: When this is specified, does axios automatically decompress the response?

JustinBeckwith: When it's specified, Google endpoints return a request with a `Content-Encoding` header set to `gzip`.  Axios will automatically decompress those requests:
https://github.com/axios/axios/blob/master/lib/adapters/http.js#L156

jinwoo: Awesome.

PR #1077, opened 2018-03-20
fix: fix download sample

### LINKED ISSUE #1060: Changes required in drive api sample
The updated sample for downloading a file using drive API needs some changes(minor). I am listing those here-
1. The scope mentioned in the current sample does not allow files to be downloaded. It allows read-only access to the metadata and not the content.
2. Currently, the file Id has been hardcoded. The file id should be passed as an argument (according to this [readme](https://github.com/google/google-api-nodejs-client/blob/master/samples/drive/README.md)).

My current environment-
googleapis @ 27.0.0
node @ 8.9.4
npm @ 5.6.0

Thanks!

JustinBeckwith:
Resolves #1060

### DIFF
--- a/samples/drive/download.js
+++ b/samples/drive/download.js
@@ -18,6 +18,7 @@ const sampleClient = require('../sampleclient');
 const fs = require('fs');
 const os = require('os');
 const uuid = require('uuid');
+const path = require('path');
 
 const drive = google.drive({
   version: 'v3',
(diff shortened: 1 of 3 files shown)


samples/drive/download.js
```diff
@@ -17,18 +17,18 @@ const {google} = require('googleapis');
...
 });
-const scopes = ['https://www.googleapis.com/auth/drive.metadata.readonly'];
-
-function downloadDoc (callback) {
-  // [START main_body]
-  const fileId = '0B7l5uajXUzaFa0x6cjJfZEkzZVE';
-  const dest = fs.createWriteStream(`${os.tmpdir()}/photo.jpg`);
 
+function runSample (fileId, callback) {
+  const filePath = `${os.tmpdir()}/${uuid.v4()}`;
```
alexander-fenster: use path.join()?

PR #1082, opened 2018-03-24
feat: type all the things

### LINKED ISSUE #881: Enable `noImplicitThis` on the tsconfig
Similar to #880, we should enable `noImplicitThis`

JustinBeckwith:
This one is a doozy.  This dramatically changes the way the generator works, but in a way that is backward compatible.  We're now generating es6 classes and interfaces for all of the APIs.  This paves the way for promise support, better d.ts support, and better documentation.  

As a bonus, resolves #881


test/test.discover.ts
```diff
@@ -35,7 +35,7 @@ describe('GoogleApis#discover', () => {
     localApis.splice(localApis.indexOf('index.js.map'), 1);
 
     localApis.forEach((name) => {
-      assert(g2[name]);
+      assert(g2[name], `API '${name}' is not available on GoogleApis`);
```
alexander-fenster: Thank you for this! I'm tired of putting `console.log` here every time the generator fails.

PR #1088, opened 2018-03-27
feat: add a gmail send sample

### LINKED ISSUE #739: Send email with UTF-8 encoded subject
How to send email with UTF-8 encoded subject and From name.

Thanks!

JustinBeckwith:
Resolves #739

### DIFF
--- a/samples/gmail/send.js
+++ b/samples/gmail/send.js
@@ -1,0 +1,81 @@
+// Copyright 2018, Google, LLC.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


samples/gmail/send.js
```diff
@@ -0,0 +1,81 @@
...
+const gmail = google.gmail({
+  version: 'v1',
+  auth: sampleClient.oAuth2Client
+});
+
+function runSample (callback) {
+  // You can use UTF-8 encoding for the subject using the method below.
+  // You can also just use a plain string if you don't need anything fancy.
+  const subject = '🤘 Hello 🤘';
+  const utf8Subject = `=?utf-8?B?${Buffer.from(subject).toString('base64')}?=`;
```
alexander-fenster: maybe `libmime.encodeWord(subject)` would be better? not sure if we want to use more dependencies in samples though

JustinBeckwith: Oh, that's a neat one :)  As you call out, if possible, I'd really like to limit the # of packages we use here.  I'd prefer to be a tad more verbose, but show the concept.

PR #1105, opened 2018-04-05
docs: use typedoc to re-generate ref docs

### LINKED ISSUE #1009: generate-docs not working
Hello,
[contributing.md](https://github.com/google/google-api-nodejs-client/blob/master/.github/CONTRIBUTING.md) file mentions about using  "npm run generate-docs" to generate documentation. But the "generate-docs" script is not present in the package.json file.

P.S: It's my first time raising any issue so please correct me if I did anything incorrectly.

Thanks!

### LINKED ISSUE #1035: Improve the documentation
I'm really sorry to say that but, well, this documentation is bad. I'm sorry to just complain. I may contribute in the meantime to fix some issues, but for now I just wanted to tell why I don't like it and get some feedback or guidance (maybe I'm just blind) and find out what you think about it.

I thought that this kind of a library from a big and respected company as Google, stating `Google's officially supported Node.js client library for accessing Google APIs.` will be easy to dive into. Unfortunately I wasted a lot of time figuring out how to work with an API client (which IMO should simplify my workflow greatly) and most likely would end up a lot better just using the services via REST APIs.

Here's why:

## ToC in the [main docs](http://google.github.io/google-api-nodejs-client/22.2.0/index.html)
The ToC there is completely broken - only the right hand side ToC works as expected.

## ToC in README
There are non-working links in the ToC in the main README, like [this](https://github.com/google/google-api-nodejs-client#migrating-to-version-25x-of-this-library). But I guess it's the only one here.

## Getting the Google Apis instance
In the examples of usage in the main README there is:
```js
var { google } = require('googleapis');
var urlshortener = google.urlshortener('v1');
```
while in the main docs:
```js
var google = require('googleapis');
var urlshortener = google.urlshortener('v1');
```
which of course won't work because there will be no `urlshortener` function (note the `{ google }` object deconstructure).

## Creating a service client
This one is really, really incomprehensible.

The docs identically (docs, readme) give this as an example:
```js
var {google} = require('googleapis');
var urlshortener = google.urlshortener('v1');
```
which is ok, but when I go further I get really confused, because there are examples like:
```js
var drive = google.urlshortener({
  version: 'v2',
  auth: oauth2Client
});
```
so I go to find some kind of reference documentation at the [main docs](http://google.github.io/google-api-nodejs-client/22.2.0/index.html) page, but there also is nothing telling that while creating a service I can invoke it with a `string` (with a version), or with a `object` (with `auth` and `version` properties). These are just in the examples... In the [Authentication section](https://github.com/google/google-api-nodejs-client#authorizing-and-authenticating) (and not in the [Create service client section](https://github.com/google/google-api-nodejs-client#create-a-service-client))

## Mysterious apirequest_1
Like.. What the hell is [this](http://google.github.io/google-api-nodejs-client/22.2.0/global.html)?

So... What do you think? Is this just me or is this documentation really unclear?

### LINKED ISSUE #1063: Readme and documentation differs on require example
On the `http://google.github.io` documentation pages this way of `request`ing the library is given:

`
const google = require('googleapis');
const tagmanager = google.tagmanager('v2');
`

I found this didn't work, and went confirm that this was the same on the readme. I found that it differs slightly:

`
var {google} = require('googleapis');
var urlshortener = google.urlshortener('v1');
`

Interestingly, the use of `const` is more modern than `var`, but the example with `var` is the one that works 😄 

It would be a good idea to update the `http://google.github.io` pages.

Thanks!

JustinBeckwith:
The previous jsdocs were waaaaaay out of date and broken.  These are already published:
http://google.github.io/google-api-nodejs-client/

Resolves #1009 
Resolves #1063 
Resolves #1035


jsdoc-conf.json
```diff
@@ -1,42 +0,0 @@
-{
```
alexander-fenster: do we still need jsdoc dependency then?

JustinBeckwith: Nope!  Pretty sure I deleted it.

sarneeh:
@JustinBeckwith All the globals on the new docs page return 404 :(

PR #1108, opened 2018-04-06
fix: Replace string-template with url-template

### LINKED ISSUE #1099: createApiRequest encodes url, resulting in 404 for pubsub endpoints
I'm trying to pull() a subscription using `pubsub.projects.subscriptions.pull()` 

```javascript
const pubsub = google.pubsub('v1');
const request = {
    subscription: 'projects/project-1235/subscriptions/my-subscription',
    auth: oAuthClient,
    resource: {
        maxMessages: 1,
        returnImmediately: true,
    },
};

pubsub.projects.subscriptions.pull(request, cb);
```
But it results in a 404. 

I looked around and apparently it is because [this line here](https://github.com/google/google-api-nodejs-client/blob/master/src/lib/apirequest.ts#L128
) encodes the URI and forward slashes become `%2F` and google doesn't recognise it. 

This is probably required for other requests so i'm not quite sure what the fix would be.

Currently I'm creating the request myself and calling it with `authClient.request()`. If anyone else is stuck with this, here's my temporary fix:

```javascript
const project = 'my-project-12345';
const subscription = 'my-subscription';

const pullReq = {
    url: `https://pubsub.googleapis.com/v1/projects/${project}/subscriptions/${subscription}:pull`,
    method: 'POST',
    paramsSerializer: params =>
        qs.stringify(params, { arrayFormat: 'repeat' }), // qs = require('qs')
    params: {
        maxMessages: 1,
        returnImmediately: true,
    },
    maxContentLength: 2147483648,
    validateStatus: status =>
        (status >= 200 && status < 300) || status === 304,
};

oAuthClient.request(pullReq).then(response => { ...
```

MennoTammens:
Fixes #1099

Path parameters are incorrectly encoded. This fix uses a different module to parse the path parameters, using the same RFC6570 format as the API Discovery Documents uses.

From the [API Discovery Service Documentation]:(https://developers.google.com/discovery/v1/using#build-compose):
> The `path` property of each method in the API uses [URI Template v04](http://code.google.com/p/uri-templates/) syntax. This property may contain variables, which are surrounded by curly braces.

In `google-api-nodejs-client` the special URI Template operators (`+`, `/` and `*`) were stripped from the path, to make the templates compatible with the `string-template` module.  
The `string-template` module should be replaced with a URI Template module.

- [x] `npm test` succeeds
- [ ] Pull request has been squashed into 1 commit
  - 1 commit with actual changes, 1 commit with regenerated files
  - I actually did! src/apis/videointelligence/v1beta1.ts is generated with wrong formatting


MennoTammens:
CLA fixed

PR #1115, opened 2018-04-11
fix: add semver ranges to package.json

JustinBeckwith:
Not quite sure how these got removed, but it's causing a lot of greenkeeper churn.


alexander-fenster:
Want to try [Renovate](https://renovateapp.com)? The main benefit is that it supports `package-lock.json` properly. You can grab the sample config from [here](https://github.com/dialogflow/dialogflow-nodejs-client-v2/blob/master/renovate.json). Also, it auto-rebases its own PRs which is useful.

JustinBeckwith:
Absolutely!  I am going to be moving towards sub-packages here very soon, so I think I'll wait to see how y'all sort that out first though :)

PR #1141, opened 2018-05-01
feat: add typing for request params

JustinBeckwith:
Scroll to the bottom for the good part 🐱


acchou:
This is exactly what I've been looking for... and was about to investigate working on :) Thanks and hopefully it can be vetted and merged soon.

JustinBeckwith:
@google/google-node-team what, nobody wants to review my 300,000 line change?

acchou:
I took a look at the generated types and have a few comments. Mostly I was reviewing the `compute` api v1.  Consider `compute.instances.insert` and the parameter type `Params$Resource$Instances$Insert`:

```typescript
export interface Params$Resource$Instances$Insert {
    auth?: string|OAuth2Client|JWT|Compute|UserRefreshClient;
    project: string;
    requestId?: string;
    sourceInstanceTemplate?: string;
    zone: string;
    resource?: Schema$Instance;
  }
```

`project` and `zone` could be set at the global or api level so should probably be optional?

The resource of type `Schema$Instance` is optional, though it appears to be required. Looking at its type:

```typescript
export interface Schema$Instance {
    ...
    /**
     * [Output Only] The CPU platform used by this instance.
     */
    cpuPlatform?: string;
```

Not sure if there's a way to separate out output-only fields that clutter this type when using it as an input. Using `Pick<T,K>` is not going to cut it for nested objects, so unfortunately doing this might require generating a whole other set of types. So probably overkill...

On the other hand, having unified input and output Schema types means making pretty much all fields need to be optional, even if the responses from the API always have certain fields. This makes using the results somewhat less convenient because it will be hard to tell where error checks are really needed. I expect to see client code littered with `!`'s and/or checks for missing fields that can't actually be missing. Splitting it into input and output instances might look more like:

```typescript
export interface Schema$Instance {
   cpuPlatform: string;   // not optional, always returned by server
   name: string;  // always returned, not optional.
   ...
}
export interface Schema$InstanceInsert {
    // cpuPlatform isn't even mentioned here because it's output-only
    name: string;   // required for creating an instance
    ...
}
export interface Params$Resource$Instances$Insert {
   resource: Schema$InstanceInsert;   // Not optional, required to create an instance
   ...
}
```

Of course this is only possible if the source API definition file has the detail required.  Maybe this is reaching for the stars... I'd be very happy with just the types you already have generated.

JustinBeckwith:
This is really great feedback.  

> project and zone could be set at the global or api level so should probably be optional?

Yeah - that's really tough.  If we go down that road, I just need to make every property optional, and do required field checking at runtime (which I need to do anyway).  The tradeoff here is that you won't really know what's required or not statically.  I could really go either way on this one 🤷‍♂️ Totally open for debate. 

> The resource of type Schema$Instance is optional, though it appears to be required. 

Sadly, this is one of those places where I'm limited by what information is available in the discovery docs I'm using to generate these APIs.  Lets use compute.instances.insert as an example:
https://www.googleapis.com/discovery/v1/apis/compute/v1/rest

The discovery doc tells me that the body of the request can take an `Instance` object, but *doesn't* tell me if it's required.  The schema object it points to also doesn't give me any information on which fields within that object are actually required either.  Sadly, this is one of those places where I just don't have enough information to do what I'd like.

JustinBeckwith:
You've changed my mind :)  I made all the params optional.  On how the discovery doc is generated...  I'm not sure!  I will find out.

acchou:
Awesome. Really looking forward to using these APIs with types.

alexander-fenster:
@acchou re: compute - we actually have another Node.js client with manually written (not generated) methods, want to check it out? [@google-cloud/compute](https://www.npmjs.com/package/@google-cloud/compute)

alexander-fenster:
(but it's plain JavaScript, no TypeScript yet while it's planned :) )

src/scripts/generator.ts
```diff
@@ -109,7 +109,13 @@ export class Generator {
   }
 
   private cleanPropertyName(prop: string) {
-    return prop.replace('@', '').replace('-', '');
+    const match = prop.match(/@|\-|\./g);
```
alexander-fenster: nit: no need to use `|` here, character class `[-@.]` is enough

src/scripts/generator.ts
```diff
@@ -109,7 +109,13 @@ export class Generator {
...
+    return match ? `'${prop}'` : prop;
+  }
+
+  private hasResourceParam(method: SchemaMethod) {
+    return method.parameters &&
+        Object.keys(method.parameters).indexOf('resource') > -1;
```
alexander-fenster: why not just `'resource' in method.parameters`?

src/templates/api-endpoint.njk
```diff
@@ -19,6 +19,7 @@
  */
 
 import { GoogleApis } from '../..';
+import {OAuth2Client, JWT, Compute, UserRefreshClient} from 'google-auth-library';
```
alexander-fenster: nit: this line and its neighbors are formatted differently: `{ GoogleApis }` but `{AxiosPromise}`

test/test.media.ts
```diff
@@ -99,7 +99,7 @@ describe('Media', () => {
     const fileName = path.join(__dirname, '../../test/fixtures/mediabody.txt');
     const fileSize = (await pify(fs.stat)(fileName)).size;
     const google = new GoogleApis();
-    const youtube = google.youtube('v3');
+    const youtube = google.youtube<youtube_v3.Youtube>('v3');
```
alexander-fenster: man, that type parameter makes things so much less readable :(

src/scripts/generator.ts
JustinBeckwith: Because I just love using `Object.keys` :) . Good catch.

test/test.media.ts
JustinBeckwith: Yeah :(  The issue here is a little complicated.  Because we offer more than one version of an API by default, I have to join the types if there are multiple.  If there's a single API, it's great.  If the multiple APIs have the same shape.. it's great!  When there are multiple versions and they're different... meh.  The only alternative I can think of here is to create a different API that references the specific typed API directly.

alexander-fenster: Yeah, maybe it's possible to make `const youtube = google.youtube_v3` work. I see one more problem here, and it's huge: assume you have `v1` which works like `const something = google.something('v1')` but then API owners publish an incompatible `v2` that, if I read your explanation correctly, breaks the existing code as it will require `const something = google.something<something_v1.Something>('v1')` and the existing code will stop working. Is my understanding correct?

alexander-fenster: Or did I understand your response wrong and it currently requires to use type parameter in all cases, even if there is just one version published?

JustinBeckwith: So - a few things.  Technically, the change to use these types was made a few commits ago.  The difference with *this* commit is that now instead of it just dictating the types of the response, it also dictates the types of the request.  

The problem you're pointing out here is very real.  There are a few options I can think of:

1. Make the generic type parameter required.  Today it defaults to `api_v1 | api_v2 | api_v3`. 

2. Make the generic type parameter default to `any`.  This would mean losing intellisense for anyone not specifically passing the type parameter.

3. Make the generic type parameter default to the latest semver version of the API.  This is *really* tricky.  But we could make our best guess of which generic type parameter to use.  The risk there is that the user chose a different version, and their type hinting is just flat out wrong. 

4. Introduce a new API for TypeScript users that doesn't rely on the version being passed in as a function argument. For example:

```js
const youtube = google.youtube.v3();
```

None of these are perfect, but I'm very interested in others thoughts :)

@google/google-node-team

acchou: @JustinBeckwith Maybe you could use [overloads](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) here? Here's a conceptual example (untested, I believe it is valid code):
```typescript
class youtube_v2 {
    _foo: number;
}
class youtube_v3 {
    _bar: number;
}
class Google {
    youtube(version: "v2"): youtube_v2;
    youtube(version: "v3"): youtube_v3;
    youtube(version: string): youtube_v3 | youtube_v2 {
        switch (version) {
            case "v2":
                return new youtube_v2();
            case "v3":
                return new youtube_v3();
            default:
                return new youtube_v3();
        }
    }
}

const g = new Google();
let v3 = g.youtube("v3"); // v3: youtube_v3
let v2 = g.youtube("v2"); // v2: youtube_v2
v3 = v2; // type error
```

alexander-fenster: OK so I don't like (1) because it looks weird and the user needs to type more. I don't like (3) because it will make a possible semver minor update (adding new version is semver minor) incompatible as the default will change. I don't really care about (2) because I don't use intellisense :) but I understand why it's important. So, (4)? (as an alternative, `google.youtube_v3()` also LGTM)

JustinBeckwith: @acchou that's a really nifty idea, and for the case where you're just passing the version string I think it would work.  I wonder if this would handle the object literal case as well?

```js
const youtube = google.youtube({
  version: "v3",
  auth: myOAuth2Client
});
```

kjin: I like what @acchou is suggesting here. Is that a possibility?

jinwoo: +1 to what @acchou suggests. One thing I like about TypeScript is it can dispatch to a different function signature depending on a string constant value of a parameter; e.g. `EventEmitter`'s `on()` or `emit()` function, or like this case.

acchou: @JustinBeckwith Indeed it does seem possible to handle object literals as well (reusing some definitions from above):
```typescript

interface GlobalOptions {
    url?: string;
    method?: string;
    baseURL?: string;
    auth?: any;
}
interface ServiceOptions extends GlobalOptions {
    version?: string;
}
interface ServiceOptionsv3 extends ServiceOptions {
    version: "v3";
}
interface ServiceOptionsv2 extends ServiceOptions {
    version: "v2";
}
class Google {
    youtube(version: "v2"): youtube_v2;
    youtube(version: "v3"): youtube_v3;
    youtube(options: ServiceOptionsv2): youtube_v2;
    youtube(options: ServiceOptionsv3): youtube_v3;
    youtube(options: string | ServiceOptions): youtube_v3 | youtube_v2 {
        let version = typeof options === "string" ? options : options.version;
        switch (version) {
            case "v2":
                return new youtube_v2();
            case "v3":
                return new youtube_v3();
            default:
                return new youtube_v3();
        }
    }
}

let o3 = g.youtube({ version: "v3" });
let o2 = g.youtube({ version: "v2" });
o3 = o2; // type error

let z3 = g.youtube({ version: "v3", auth: "foo" });
let z2 = g.youtube({ version: "v2", auth: "bar" });
z3 = z2; // type error
```

acchou:
Note that overloading with an exact object literal doesn't work:
```typescript
    youtube(options: { version: "v2" }): youtube_v2;
    youtube(options: { version: "v3" }): youtube_v3;
```
This will cause  type error in the z3 and z2 cases above, because the type is hard-coded to that specific object literal without allowing for any of the other keys in `GlobalOptions`. That's why we need `ServiceOptionsv2` and `ServiceOptionsv3`.

... and that means we may need to add two more overloads:
```typescript
    youtube(version: string): youtube_v3 | youtube_v2;
    youtube(options: ServiceOptions): youtube_v3 | youtube_v2;
```
It's debatable if these are needed. Without them, bugs like this are caught by the type checker:
```typescript
g.youtube("xyz");  // type error
g.youtube({ version: "xyz" }); // type error
```
The two extra overloads only seem to be needed if you expect to have a dynamically generated `version`. This could happen in a developer tool, for example. In that case, the code can cast the version or option to `any`.

My preference would be to leave the non-explicit version overloads out. It preserves a better experience and type checking for 99%+ of the users of the API.

test/test.media.ts
acchou: Note that overloading with an exact object literal doesn't work:
```typescript
    youtube(options: { version: "v2" }): youtube_v2;
    youtube(options: { version: "v3" }): youtube_v3;
```
This will cause  type error in the z3 and z2 cases above, because the type is hard-coded to that specific object literal without allowing for any of the other keys in `GlobalOptions`. That's why we need `ServiceOptionsv2` and `ServiceOptionsv3`.

... and that means we may need to add two more overloads:
```typescript
    youtube(version: string): youtube_v3 | youtube_v2;
    youtube(options: ServiceOptions): youtube_v3 | youtube_v2;
```
It's debatable if these are needed. Without them, bugs like this are caught by the type checker:
```typescript
g.youtube("xyz");  // type error
g.youtube({ version: "xyz" }); // type error
```
The two extra overloads only seem to be needed if you expect to have a dynamically generated `version`. This could happen in a developer tool, for example. In that case, the code can cast the version or option to `any`.

My preference would be to leave the non-explicit version overloads out. It preserves a better experience and type checking for 99%+ of the users of the API.

JustinBeckwith:
Thanks for the vibrant discussion and ideas folks!  I think we got it :)  Take a look and let me know what you think.

acchou:
Looks good to me. Thanks for taking this one step further to make it more convenient for clients of the api.

PR #1149, opened 2018-05-04
fix: use `requestBody` instead of `resource`

### LINKED ISSUE #721: Warn about renaming `resource` method parameter names
Method parameters named `resource` are renamed to `resource_` to avoid name collision with request body fields, but this not obviously documented, easily overlooked, and currently provides no helpful warning on failure. See https://github.com/google/google-api-nodejs-client/issues/720.

JustinBeckwith:
This is a backwards compatible change that changes the `resource` special case name to `requestBody`.  Doing this for 2 reasons:

1. It's fairly common to have a `resource` parameter required independent of the body.  We have a janky thing right now where we force clients to use `resource_` in this case, which is neither obvious nor well documented.

2. `requestBody` does a better job of explaining what this object actually is and what it does. 

Resolves #721

### DIFF
--- a/src/lib/apirequest.ts
+++ b/src/lib/apirequest.ts
@@ -82,7 +82,27 @@ async function createAPIRequestAsync<T>(parameters: APIRequestParams) {
   );
 
   const media = params.media || {};
-  const resource = params.resource;
+
+  /**
+   * In a previous version of this API, the request body was stuffed in a field
+   * named `resource`.  This caused lots of problems, because it's not uncommon
+   * to have an actual named parameter required which is also named `resource`.
+   * This mean that users would have to use `resource_` in those cases, which
+   * pretty much nobody figures out on their own. The request body is now
+   * documented as being in the `requestBody` property, but we also need to keep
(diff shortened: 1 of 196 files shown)


JustinBeckwith:
*gentle ping* @alexander-fenster

src/lib/apirequest.ts
```diff
@@ -82,15 +82,34 @@ async function createAPIRequestAsync<T>(parameters: APIRequestParams) {
...
+   * In a previous version of this API, the request body was stuffed in a field
+   * named `resource`.  This caused lots of problems, because it's not uncommon
+   * to have an actual named parameter required which is also named `resource`.
+   * This mean that users would have to use `resource_` in those cases, which
+   * pretty much nobody figures out on their own. The request body is now
+   * documented as being in the `requestBody` property, but we also need to keep
+   * using `resource` for reasons of back-compat. Cases that need to be covered
+   * here:
+   * - user provides just a `resource` with a request body
+   * - user provides both a `resource` and a `resource_`
```
alexander-fenster: The code below does not handle this case at all (probably because it's not really needed to handle it), why mention it here?

PR #1155, opened 2018-05-08
docs: remove caveat about TypeScript docs

### LINKED ISSUE #1154: Update README re: TypeScript
The README refers to the need for using generic types instead of relying on the overloading provided in PR #1141

```typescript
import {google, drive_v3} from 'googleapis';
const drive = google.drive<drive_v3.Drive>('v3');
```
Probably worth updating this before the next release.

JustinBeckwith:
resolves #1154

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,14 +465,7 @@ const apis = google.getSupportedAPIs();
 This will return an object with the API name as object property names, and an array of version strings as the object values;
 
 ### TypeScript
-This library is written in TypeScript, and provides types out of the box. Due to the shape of the API, in some cases you may need to use generics to get the appropriate types.  This is only an issue for APIs that include multiple versions.  For example, to use the Drive v3 API:
-
-```ts
-import {google, drive_v3} from 'googleapis';
-const drive = google.drive<drive_v3.Drive>('v3');
-```
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -465,14 +465,7 @@ const apis = google.getSupportedAPIs();
...
 ### TypeScript
-This library is written in TypeScript, and provides types out of the box. Due to the shape of the API, in some cases you may need to use generics to get the appropriate types.  This is only an issue for APIs that include multiple versions.  For example, to use the Drive v3 API:
-
-```ts
-import {google, drive_v3} from 'googleapis';
-const drive = google.drive<drive_v3.Drive>('v3');
-```
-
-All classes and interfaces generated for each API are exported under the `apiName_version` namespace.
+This library is written in TypeScript, and provides types out of the box. All classes and interfaces generated for each API are exported under the `apiName_version` namespace.
```
ofrobots: Is `apiName_version` the literal namespace? It sounds like it. If it is not a literal namespace, can you something like: '..exported under the `${apiName}_version` namespaces, e.g. `LullabyAPI_v99`'.

JustinBeckwith: Good feedback!  Updated.

PR #1158, opened 2018-05-09
Remove curly-braces from google object declaration

rainabba:
Breaks the sample because the 'drive' object doesn't exist later.

```js
TypeError: Cannot read property 'auth' of undefined
    at Object.<anonymous> (google_node_api_quickstart.js:28:27)
```

- [ ] `npm test` succeeds
- [ ] Pull request has been squashed into 1 commit
- [ ] Appropriate changes to readme/docs are included in PR

### DIFF
--- a/samples/drive/quickstart.js
+++ b/samples/drive/quickstart.js
@@ -14,7 +14,7 @@
 'use strict';
 
 // [START main_body]
-const {google} = require('googleapis');
+const google = require('googleapis');
 const express = require('express');
 const opn = require('opn');
 const path = require('path');


JustinBeckwith:
Greetings, and thanks for the PR!  While what you have written here would work for versions below v25, it would not work for versions greater than 25 :)   Those curly braces are very much there on purpose.  

Can you share which version of the library you're using?

rainabba:
I don't question the truth of that statement
I'm running node 8.11.1 and using v25 since it works. I'd have expected the most current docs to work with the most current api (v29), but it breaks for different reasons that I have yet to understand.

@JustinBeckwith I'd be happy to make myself available to discuss in real time (phone/hangout/etc), but I'm not kidding when I say that I've been at this for more than a year and each time I've spent days trying to get things going. The biggest problem is that there are *SO* many docs for Google and not a single one works as its listed. I think there are too many assumptions that aren't obvious to me because this has been going on since node 4.x. I'd be happy to help further here if my time is used well and I don't think it will be using text-chat here or else we wouldn't be having this conversation.

rainabba:
Naturally after confirming v25 with and without braces (only works without), I then confirmed v29 WITH the braces just to be sure, and it works. No doubt something else (or many) were tripping me up though now that I've got it working, I can't imagine what it was.

I'm going to try and slip back into the shadows now :)

JustinBeckwith:
No problems!  If you do run into a sample or doc that doesn't "just work" - I do very much want to hear about it.  It's true that this library didn't get a lot of love for a while, but that's no longer the case :)

Let me know how to make it better, and I'm more than happy to help. Thanks!

rainabba:
There are quite a few helpers out there to make auth "easier". For whatever reason, none of them worked for me (and I take responsibility for that) so once I got things working, I decided to share. This isn't published to NPM yet (need to add more tests and polish a bit more), but it will be there soon.

https://github.com/rainabba/oauth-token-generator-google

PR #1161, opened 2018-05-10
fix: use reject handler instead of catch for callbacks

### LINKED ISSUE #1159: Unexpected behavior when callback throws an error
Consider the following small modification of the UrlShortener example in the README:

```js
urlshortener.url.get(params, (err, res) => {
  if (err) return console.error('The API returned an error: ' + err);
  console.log(`Long url is ${res.data.longUrl}`);
  throw new Error('This error is not from the API');
});
```

### Expected behavior

The final error is bubbled up the stack and logged in the console.

### Actual behavior

The final error is funneled back into my callback and treated as an API error:

```
Long url is https://www.google.com/
The API returned an error: Error: This error is not from the API
```

This seems to be a side effect of [setting the callback as the `catch()` handler for the promise](https://github.com/google/google-api-nodejs-client/blob/a783911c95a83fefb08e5fb49ee9817af5cbdab4/src/lib/apirequest.ts#L64). The catch executes not only for errors resulting from the API call but also for any errors thrown in the `then()` handler for a successful call.

I'm not sure what Promise interface Axios supports, but I think the solution would be to set the callback as the error handler for a specific `then()`, instead of as a `catch()`:

```js
createAPIRequestAsync<T>(parameters)
        .then(r => callback(null, r),
              e => callback(e));
```

JustinBeckwith:
Plus a bonus sample fix. Resolves #1159

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "googleapis",
-  "version": "29.0.0",
+  "version": "30.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/samples/urlshortener/urlshortener.js
+++ b/samples/urlshortener/urlshortener.js
@@ -18,7 +18,11 @@ const path = require('path');
 const nconf = require('nconf');
 
 nconf.argv().env().file(path.join(__dirname, '../config.json'));
-const urlshortener = google.urlshortener('v1', nconf.get('api_key'));
+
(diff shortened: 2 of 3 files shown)


JustinBeckwith:
@ofrobots @alexander-fenster *gentle ping*

alexander-fenster:
why `package-lock.json`?

JustinBeckwith:
Thanks!  It was just picking up the version change.

PR #1163, opened 2018-05-10
fix: drop support for node.js 4

JustinBeckwith:
Now that node.js 4 is EOL, I would like to explicitly drop support for it here.  Thoughts / opinions welcomed :)


ofrobots:
I'm +1 for this particular library. This should probably get approved by @alexander-fenster.

Issue #806, opened 2017-09-06
Publish packages for each individual API.


lukesneeringer:
This is going to be done in several steps:
- [x] Re-organize the directory structure to map to individual modules (#1167)
- [x] Setup and publish a `googleapis-common` npm module that just contains shared bits that will be used by all packages
- [x] Generate a package.json for all of the individual APIs
- [ ] Script the `npm publish` process so that each individual package is published along with the larger meta-package

These steps are of course subject to change as we discover things through the process :) 

--- UPDATE ---
We're down to the last step of actually publishing the individual modules.  @alexander-fenster, @jkwlui, and @JustinBeckwith last got together to chat about this, and came to a few conclusions:

### State of the world today
- We decide when to run the generator in a somewhat ad-hoc fashion.  As someone asks for a release, we cut it. 
- We run the generator by having a developer clone the repo, run `npm run generate`, and then submitting a PR with like 1344247584 changes. 
- After checking in that PR, someone has to submit a new release with releasetool.  This will create a new tag.  It is almost always semver major (see the fact that we're on v38).
- The generator is somewhat monolithic.  It works by blasting away the `src/apis` directory, and re-creating every API from scratch every time.  This makes it easy to detect new APIs, and easy to detect removed APIs.  

### Where we need to be
- We want a bot that runs the generator(s) nightly.  We should use `synthtool` for this.
- The bot should submit *many* PRs - one for each API removed, deleted, or modified. 
- The individual commits *could* use the [conventional commit scope](https://www.conventionalcommits.org/en/v1.0.0-beta.3/#commit-message-with-scope) as a way to signal which API/package this change applies to
- Some combination of `releasetool` and `semantic-release` could be used to cut individual releases.  A scoped tag would be used for each package release from the mono-repo.
- The risk of this approach is that the number of tags could get out of control, *fast*.   
- Alternatively, we could approach this similar to the way the gRPC generated libraries work.  We create a new GitHub repository for each API, and break the output out of this repo and into the individual package repos. 

There's still a lot to figure out, but I suspect @bcoe is gonna love this problem.

nicoabie:
@lukesneeringer #1167 is already merged so it can be marked as done I believe.
Are tasks created for all the pending steps in order to have this feature implemented?

JustinBeckwith:
[...]

PR #1169, opened 2018-05-14
refactor: isolate the googleapis-common module

JustinBeckwith:
This change entirely isolates the shared code required for all APIs from the rest of the module.  It still includes these files in the overall build, but also introduces a lower build specifically for the new module. This is another step on the way to #806.


.circleci/config.yml
```diff
@@ -64,12 +64,24 @@ workflows:
...
+            - node9
+            - node10
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: /^v[\d.]+$/
       - publish_npm:
           requires:
             - node6
```
alexander-fenster: no need to have these `node*` dependencies if it depends on `publish_shared`. It just makes the graph look weird. Of if you want parallel publishing, then remove the `publish_shared` dependency.

.circleci/config.yml
```diff
@@ -107,6 +119,25 @@ jobs:
...
+      - image: node:8
+        user: node
+    steps:
+      - checkout
+      - run:
+          name: Switch to the common directory
+          command: cd src/shared
+      - run:
+          name: Install modules and dependencies.
+          command: npm install --unsafe-perm
```
alexander-fenster: no need to use `--unsafe-perm` since we run as non-root account (here and everywhere)

src/shared/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "googleapis-common",
+  "version": "0.0.1",
```
alexander-fenster: 0.1.0 is much better as 0.0.1 prevents any patch updates (effectively pins the version).
```
> semver.satisfies("0.0.1", "^0.0.1")
true
> semver.satisfies("0.0.2", "^0.0.1")
false
```

JustinBeckwith:
Thanks @alexander-fenster - all fixed!

alexander-fenster:
Since we'll eventually have hundreds of npm modules published from this one repo, we'll need to rethink the current model of tagging and releasing based on the tag. I have no idea how exactly it should be done though (probably worth opening a separate issue to discuss future versioning).

JustinBeckwith:
One step at a time :)

ofrobots:
Oh, and next time, can you please sequester autogenerated code changes into a separate commit. Reviewing changes is a lot harder when manual changes are intermixed with gobs of generated changes.

JustinBeckwith:
Great questions! 

> Is the idea that googleapis-common would be published as a standalone package, and that this repo is now a monorepo?

YES!

> If so, let's use an established pattern of shared sub-module being in packages/googleapis-common subdirectory rather than ad-hoc src/shared.

I'm 100% on board with that.  I am trying to make these changes incrementally.  Today, the top level build/test/publish commands will continue to work, exactly as they used to.  I'd like to propose landing this first, and working towards the properly named packages.  Mostly, I'm worried about breaking the top level build.

> If monorepos are the way to go, have you considered using https://github.com/lerna/lerna?

Yup!  This is a point in time thing.  Once I get all the sub-packages working independently, we will remove the top level build and use something like lerna to orchestrate the publish and linking. 

> Oh, and next time, can you please sequester autogenerated code changes into a separate commit. Reviewing changes is a lot harder when manual changes are intermixed with gobs of generated changes.

Sorry!  Will do next time.  

shame

PR #1187, opened 2018-05-23
fix: cd before publishing

### LINKED ISSUE #1186: 31.0.0 has empty NPM package
Hi!

Just updated to 31.0.0, and the package seems to be mostly empty.

Only three files appear in the root of the package.

> README.md
> node_modules
> package.json

Indeed, the build folder is not in the package.
https://registry.npmjs.org/googleapis/-/googleapis-31.0.0.tgz

Thanks!

JustinBeckwith:
😐 resolves #1186

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,17 +122,14 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Switch to the common directory
-          command: cd src/shared
-      - run:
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
@ofrobots I totally agree.  This isn't the first time I've borked a build because we don't have a way to test it.

PR #1197, opened 2018-05-29
chore: run the generator, fixing samples

fhinkel:
This fixes the samples from `var google = require()` to
`const {google} = require()`.

But `npm test` fails. Can anybody else run `make generate` and see if they have better luck? 

```
  101 passing (12s)
  1 failing

  1) GoogleApis#discover
       should generate all apis:
     AssertionError [ERR_ASSERTION]: API 'resourceviews' is not available on GoogleApis
      at localApis.forEach (test/test.discover.ts:46:13)
      at Array.forEach (<anonymous>)
      at Context.it (test/test.discover.ts:45:15)

```


fhinkel:
I started with a clean build and turned it Off and On again™️. All tests passing now.  Please take a look.

fhinkel:
Wait, is your generator better than my generator? :smile: 
Thanks for taking care of this.

PR #1207, opened 2018-06-05
docs: fix samples link in README.md

### LINKED ISSUE #1196: Samples link is broken
There is a page http://google.github.io/google-api-nodejs-client/#samples that in turn has a link http://google.github.io/google-api-nodejs-client/samples/ which is broken. It probably should lead to https://github.com/google/google-api-nodejs-client/tree/master/samples

kinwa91:
Fixes #1196

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ runSample().catch(console.error);
 ```
 
 ### Samples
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
We are trying to use [convention commit](https://conventionalcommits.org/) style commit messages to make it easier to make change logs.  I updated this one, but from here on out please use the prefix!

kinwa91:
Thanks for fixing it!

PR #1214, opened 2018-06-14
feat: generate package.json and README per package

JustinBeckwith:
Inches us closer to shipping individual packages 🌮


src/apis/abusiveexperiencereport/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@google/abusiveexperiencereport",
```
alexander-fenster: are we sure that names for all libraries are unique on npm? maybe use some other npm org?

JustinBeckwith: They're all scoped under @google, so I'm pretty sure :)

PR #1239, opened 2018-07-17
Sheets: Make "batchGet"/range option of type string[] instead of string

arthurjdam:
Not sure why the documentation doesn't agree, but there does not seem to be any way of sending a singular string as batchGet's "ranges" option 
Example: 
`ranges: "TABONE!A1:B,TABTWO!A2:B"` will fail
`ranges: ["TABONE!A1:B", "TABTWO!A2:B"]` works just fine

- [x] `npm test` succeeds


JustinBeckwith:
Thank you so much for the PR!  Sadly, this is an auto-generated file.  Any fixes required will need to be made on the generator.  We're tracking this over in https://github.com/google/google-api-nodejs-client/issues/1266 :)

PR #1240, opened 2018-07-17
fix: cannot read property of undefined in sampleclient.js line 42, li…

nguyendviet:
fix: cannot read property of undefined in sampleclient.js line 42, line 32 .web to .installed

oauth2.keys.json when downloaded looks like this: 
{"installed":{
  "client_id":"",
  "project_id": "",
  ...
  "redirect_uris":[...]}
}

It's also mentioned here: https://github.com/gsuitedevs/node-samples/blob/master/drive/quickstart/index.js

- [x] `npm test` succeeds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.49",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.17.5",
-        "source-map": "^0.5.0",
-        "trim-right": "^1.0.1"
+        "jsesc": "2.5.1",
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
Thanks for the PR!  As it would turn out, the `.web` vs `.installed` thing is really an artifact of what kind of service account credential you create in the developer console.  If you choose a "web" application, the key uses web.  Instead of this - would you be willing to change it to something like:

```js
const keyFile = require(keyPath);
keys = keyFile.installed || keyFile.web;
```

nguyendviet:
Thanks, Justin. I'll close this one and create another pull request following your suggestion.

PR #1241, opened 2018-07-17
fix: cannot read property of undefined in sampleclient.js line 42

nguyendviet:
fix: cannot read property of undefined in sampleclient.js line 42

Added: line 32
```
const keyFile = require(keyPath);
keys = keyFile.installed || keyFile.web;
```

oauth2.keys.json when downloaded looks like this:
{"installed":{
"client_id":"",
"project_id": "",
...
"redirect_uris":[...]}
}

It's also mentioned here: https://github.com/gsuitedevs/node-samples/blob/master/drive/quickstart/index.js

- [x] `npm test` succeeds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.49",
-        "jsesc": "^2.5.1",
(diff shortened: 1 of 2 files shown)


nguyendviet:
I certainly can. Sorry about that.

PR #1285, opened 2018-08-12
fix: add better error handling around sample client and redirect URIs

### LINKED ISSUE #1272: Using query example, and sampleclient - can't get to the point of server creation
Hello,

Just downloaded the [query example here](https://github.com/google/google-api-nodejs-client/blob/master/samples/webmasters/query.js), installed the necessary packages (open, server-destroy, fs, googleapis, path) and while using [sampleclient](https://github.com/google/google-api-nodejs-client/blob/master/samples/sampleclient.js) found that the code runs only up to the `this.oAuth2Client.generateAuthUrl` call. In that sense the server doesn't start and I seem to be stuck at this page:

image
What do I do with this code? There's no prompt in the console.

The only changes I've made are using `oauth.keys.json` with the API info, following this format (a direct download from the API Console):
```
{
  "installed": {
    "client_id": "CLIENT_ID,
    "project_id": "PROJECT_ID,
    "auth_uri": "AUTH_UTI",
    "token_uri": "TOKEN_URI,
    "auth_provider_x509_cert_url": "CERT_URL",
    "client_secret": "CLIENT_SECRET",
    "redirect_uris": ["urn:ietf:wg:oauth:2.0:oob", "http://localhost"]
  }
}
```
And setting `siteUrl` to one of our websites inside the query example.

Those two are the only changes. I would appreciate any guidance given.

Thanks

JustinBeckwith:
Fixes #1272


samples/sampleclient.js
```diff
@@ -33,15 +33,35 @@ if (fs.existsSync(keyPath)) {
...
   constructor (options) {
     this._options = options || { scopes: [] };
 
+    // validate the redirectUri.  This is a frequent cause of confusion.
+    if (!keys.redirect_uris || keys.redirect_uris.length === 0) {
+      throw new Error(invalidRedirectUri);
+    }
+    const redirectUri = keys.redirect_uris[keys.redirect_uris.length - 1];
+    const parts = url.parse(redirectUri, false);
+    if (redirectUri.length > 0 && parts.port !== '3000' || parts.hostname !== 'localhost' || parts.path !== '/oauth2callback') {
```
crwilcox: This seems a bit restrictive. Don't we just need to make sure there is a redirect URI?

PR #1313, opened 2018-08-30
fix: upgrade to the latest google-auth-library

### LINKED ISSUE #998: Can't use proxy to do JWT authentication with googleapis
I have a NodeJS API working behind a corporate proxy and I have been trying to do authentication with Google with their NodeJS client: 

```js
    const google = require('googleapis');

    function getServiceAccountInfo() {
      return {
        type: 'service_account',
        project_id: 'XXXXX',
        private_key_id: 'XXXXXX',
        private_key: 'XXXX',
        client_email: 'XXXX',
        client_id: 'XXXX',
        auth_uri: 'XXXX',
        token_uri: 'XXXX',
        auth_provider_x509_cert_url: 'XXXX',
        client_x509_cert_url: 'XXXXX'
      };
    }

    const SCOPES = 'https://www.googleapis.com/auth/firebase.messaging';

    let proxy2 = {
      host: 'proxy.hkcsl.net',
      port: 8080
    };

    const proxy3 = {
      proxy: 'http://172.25.2.6:8080'
    }

    const proxy4 = {
      proxy: {
        host: '172.25.2.6',
        port: 8080,
        auth: {
          username: '',
          password: '',
        },
      }
    }

    process.env.HTTPS_PROXY = 'https://172.25.2.6:8080';
    process.env.https_proxy = 'https://172.25.2.6:8080';
    process.env.HTTP_PROXY = 'http://172.25.2.6:8080';
    process.env.http_proxy = 'http://172.25.2.6:8080';
    google.options({
      proxy: proxy4.proxy
    });

    const key = getServiceAccountInfo();
    const jwtClient = new google.auth.JWT(
      key.client_email,
      undefined, // null,
      key.private_key,
      SCOPES,
      undefined, // null
    );
    jwtClient.authorize(function(err, tokens) {
      if (err) {
        console.error(err);
        return;
      }
      console.log(tokens.access_token);
    });
```

However, no matter how I configure the proxy option, I still either get a timeout error or something like

> ERROR  Error: write EPROTO 101057795:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:openssl\ssl\s23_clnt.c:827:

Some people managed to fix it by changing the port to 443 but it doesn't apply to my case since that port is not available to me inside the server. And I've already referenced on the discussions and solutions proposed here:

[StackOverflow: node-request - Getting error “SSL23_GET_SERVER_HELLO:unknown protocol”
][1]

[googleapis: There is no facility for passing a proxy to the oauth2client][2]

[Axios: Request to HTTPS with HTTP proxy fails][3]

[Using jwtClient.authenticate() behind a proxy results in ETIMEDOUT error][4]

It also seems like Vue is also encountering a similar [problem][5] yesterday, they changed from using axios to using request instead. Is there a similar workaround for googleapis too?

Note that I have also asked this question on StackOverflow (https://stackoverflow.com/questions/48766209/cant-use-proxy-to-do-jwt-authentication-with-googleapis).

  [1]: https://stackoverflow.com/questions/15421050/node-request-getting-error-ssl23-get-server-hellounknown-protocol
  [2]: https://github.com/google/google-api-nodejs-client/issues/336
  [3]: https://github.com/axios/axios/issues/925
  [4]: https://github.com/google/google-api-nodejs-client/issues/705
  [5]: https://github.com/vuejs/vue-cli/issues/785

JustinBeckwith:
Resolves #998


JustinBeckwith:
gentle ping @alexander-fenster

PR #1321, opened 2018-09-05
Use synth templating and update CI

JustinBeckwith:


synth.py
```diff
@@ -12,25 +12,25 @@
...
 
+repository_url = "https://github.com/google/google-api-nodejs-client.git"
 log.debug(f"Cloning {repository_url}.")
 repository = git.clone(repository_url, depth=1)
-
 log.debug("Installing dependencies.")
 shell.run("npm install".split(), cwd=repository)
-
 log.debug("Generating all libraries.")
 shell.run("make generate".split(), cwd=repository)
```
kinwa91: `s.copy(repository / "src")`

.appveyor.yml
```diff
@@ -4,9 +4,9 @@ environment:
 
 install:
   - ps: Install-Product node $env:nodejs_version
-  - ps: (new-object net.webclient).DownloadFile('https://github.com/mbuilov/gnumake-windows/raw/master/gnumake-4.2.1-x64.exe', 'C:\projects\google-api-nodejs-client\make.exe')
```
kinwa91: It's probably not okay to remove this - it installs GNU `make`.

`make` is required for `npm test` which runs `make test ..`

.github/CONTRIBUTING.md
```diff
@@ -1,66 +1,53 @@
-# Contributing
```
kinwa91: We'll lose instructions for generation if we overwrite this with the one from synthtool.

Perhaps exclude it from synth.py:

```python
s.copy(templates, excludes=[".github/CONTRIBUTING.md"])
```

.appveyor.yml
JustinBeckwith: Fixed this by getting rid of the makefile and dropping the requirement on make 🙃

PR #1338, opened 2018-09-14
Fix error handling in batchGet sample

### LINKED ISSUE #1336: analytics report - batch - not displaying error properly
Hello,

While trying the report API, I encounter an issue where I malformed the query. Unfortunately I was not able to see it. I discover that the `try catch` statement was catching well but the `console.error` was not called correctly.

May I suggest to change the example to `console.error(err)` instead of `console.error`

#### Environment details

  - OS: macOS High Sierra
  - Node.js version: 8.9.1
  - npm version: 6.1.0
  - `googleapis` version: 33.0.0

#### Steps to reproduce

  1. run samples/analyticsReporting/batchGet.js
  2. Do not see any error

there: https://github.com/googleapis/google-api-nodejs-client/blob/master/samples/analyticsReporting/batchGet.js#L55

it should be `console.error(e)` instead of `console.error`

Thanks!

JustinBeckwith:
Fixes #1336

### DIFF
--- a/samples/analyticsReporting/batchGet.js
+++ b/samples/analyticsReporting/batchGet.js
@@ -52,7 +52,7 @@ if (module === require.main) {
   const scopes = ['https://www.googleapis.com/auth/analytics'];
   sampleClient.authenticate(scopes)
     .then(c => runSample())
-    .catch(e => console.error);
+    .catch(console.error);
 }
 
 // export functions for testing purposes


samples/analyticsReporting/batchGet.js
```diff
@@ -52,7 +52,7 @@ if (module === require.main) {
   const scopes = ['https://www.googleapis.com/auth/analytics'];
   sampleClient.authenticate(scopes)
     .then(c => runSample())
-    .catch(e => console.error);
+    .catch(console.error);
```
kinwa91: 😅

PR #1377, opened 2018-10-04
Allow use of --discovery-url parameter

sgarciac:
As it is now, passing any parameter makes the generate script treat parameters as individual service definition URLs, even when the --discovery-url parameter is present.


sgarciac:
I signed it!

PR #1390, opened 2018-10-16
Fix for #1374 Generator missing standard parameters

### LINKED ISSUE #1374: drive.files.list 'fields' parameter is missing
The `drive.files.list`'s `fields` parameter is missing in the Drive V3 d.ts.

#### Environment details

  - OS: OSX
  - Node.js version: v9.4.0
  - npm version: 5.6.0
  - `googleapis` version: 34.0.0

#### Steps to reproduce

  1. Use `drive.file.list` with a `fields` param.

```ts
  const res = await drive.files.list({
    pageSize: 50,
    fields: 'nextPageToken, files(id, name)',
    q: 'mimeType="application/vnd.google-apps.script"',
  });
```
  2. Notice that the `fields` param is not supported.

Perhaps there's a new way of using this param, but I don't see it in the `v3.d.ts`.

---

### Docs

- https://developers.google.com/drive/api/v3/search-parameters#using_client_libraries

### Drive V3 discovery:
- https://developers.google.com/discovery/v1/reference/apis/getRest
  - drive, v3
- Notice `parameters.fields` string:
```js
  "fields": {
   "type": "string",
   "description": "Selector specifying which fields to include in a partial response.",
   "location": "query"
  },
```

ajaaym:
Fixes #1374  (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass

### DIFF
--- a/src/generator/templates/api-endpoint.njk
+++ b/src/generator/templates/api-endpoint.njk
@@ -83,6 +83,15 @@ export class {{ Name }} {
     return this.root;
   }
 
+interface StandardParameters {
+  {% for pname, p in api.parameters|dictsort %}
+  /**
+    * {{ p.description|oneLine|cleanComments|safe }}
+    */
+  {{ pname|getSafeParamName }}?: {{ p.type }};
+  {% endfor %}
+}
+
 {% if api.methods %}
   {% for mname, m in api.methods|dictsort %}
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
Thanks for the fix!

PR #1451, opened 2018-11-19
feat: Add support for repeated fields

### LINKED ISSUE #1266: Wrong type for /v1/people/getBatchGet parameter `resourceNames`. Should accept `string[]`
As stated in title, `resourceNames` should accept `string[]` as well as `string`:

https://github.com/google/google-api-nodejs-client/blob/master/src/apis/people/v1.ts#L2399

Detailed discussion: https://stackoverflow.com/questions/51588465/is-resourcenames-comma-separated-in-google-people-getbatchget?noredirect=1#comment90144867_51589512

Thanks!

ajaaym:
Fixes #1266  Adds support for repeated fields

- [x] Tests and linter pass

### DIFF
--- a/src/generator/templates/api-endpoint.njk
+++ b/src/generator/templates/api-endpoint.njk
@@ -45,7 +45,7 @@ interface StandardParameters {
   /**
     * {{ p.description|oneLine|cleanComments|safe }}
     */
-  {{ pname|cleanPropertyName|safe }}?: {{ p.type }};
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
Thanks! Generally LGTM - starting to think we should add something to the system tests to run the generator though.  Can you verify you ran `npm run generate`, and that the resulting output passes all the tests?

ajaaym:
@JustinBeckwith Yeah I ran below command and all runs fine.
npm run generate
npm run test
npm run system-test
npm run samples-test

I agree that we should have something to system tests to run the generator. Will create another PR for this.

PR #1508, opened 2018-12-14
Update append.js

Edwardzyc:
Looks like append takes resource instead of requestBody


Edwardzyc:
I signed it!

JustinBeckwith:
Greetings, and thanks for the PR!  This change shouldn't be needed - `resourceBody` should work :)  Are you running into issues? What version of the module are you using?

PR #1517, opened 2018-12-24
Added a sample for the Calendar API

### LINKED ISSUE #1516: Add Node.js Samples for Calendars API
The [Google Calendar Documentation](https://developers.google.com/calendar/v3/reference/calendarList/list?apix_params=%7B%7D) has poor support for Node.js. There are also *no* samples included in this repository.

While some examples exist (e.g., the [QuickStart](https://developers.google.com/calendar/quickstart/nodejs)), these can be extended to provide greater visibility into *how* these APIs can be used with Node.js.

stephencweiss:
This is a basic example of the Calendar API that uses both the CalendarList and Events APIs.

Fixes #1516. My goal with this example is to make it easier for other people to get up to speed and quickly extend the [Node.JS Quickstart](https://developers.google.com/calendar/quickstart/nodejs) example.

- [ ] Tests and linter pass


stephencweiss:
I've signed the CLA!

samples/calendar/index.js
```diff
@@ -0,0 +1,143 @@
+
```
JustinBeckwith: We need to have a license header in all of these.  You can use this as an example:
https://github.com/googleapis/google-api-nodejs-client/blob/master/samples/drive/list.js

samples/calendar/package.json
```diff
@@ -0,0 +1,15 @@
+{
```
JustinBeckwith: We generally don't include a separate `package.json` for individual samples like this, and rely on the higher level one.  I'm totally open to the idea, but for now I'd like to drop this for the sake of consistency.

samples/calendar/index.js
JustinBeckwith: Instead of `index.js`, can we name this `quickstart.js`?

samples/calendar/index.js
```diff
@@ -0,0 +1,143 @@
...
+
+const fs = require('fs');
+const readline = require('readline');
```
JustinBeckwith: So for most of the OAuth2 workflow examples, we use this `sampleClient.js` thingy that abstracts most of the goo away.  Could you check out the drive sample, and make it mostly match that approach?  It's nice, because it hides away auth for the individual interesting sample you're adding. 
https://github.com/googleapis/google-api-nodejs-client/blob/master/samples/drive/list.js

samples/calendar/index.js
```diff
@@ -0,0 +1,143 @@
...
+// If modifying these scopes, delete token.json.
+const SCOPES = ['https://www.googleapis.com/auth/calendar.readonly'];
+// The file token.json stores the user's access and refresh tokens, and is
+// created automatically when the authorization flow completes for the first
+// time.
+const TOKEN_PATH = 'token.json';
+
+// Load client secrets from a local file.
+fs.readFile('credentials.json', (err, content) => {
+  if (err) {
```
JustinBeckwith: A lot of this will be covered by making it consistent with the [drive sample](https://github.com/googleapis/google-api-nodejs-client/blob/master/samples/drive/list.js), but in the general sense we want samples to use `async/await` style syntax over callbacksies.

stephencweiss:
@JustinBeckwith Thanks for the feedback! I'll get working on the changes requested. Cheers!

JustinBeckwith:
@grant got it.  I think we're going to go ahead and close this one out for now.  In the future, we may choose to move the calendar samples to this repo, but we don't want to try and maintain two :)

PR #1520, opened 2018-12-28
build: check broken links in generated docs

kinwa91:


JustinBeckwith:
This is going to be hella complicated.

PR #1535, opened 2019-01-07
DO NOT MERGE: feat: make it available in browser

alexander-fenster:
This will not be used this way since we need to split it into smaller packages, but I will play in this branch to double check everything works as expected.


webpack.config.js
```diff
@@ -0,0 +1,70 @@
...
+    crypto: 'empty',
+  },
+  module: {
+    rules: [
+      {
+        test: /src\/crypto\/node\/crypto/,
+        use: 'null-loader',
+      },
+      {
+        test: /node_modules\/https-proxy-agent\//,
```
JustinBeckwith: It sucks that you need to call this out both here and in google-auth-library.  Is there a way to extend the config here for users so they don't have to suffer this?

alexander-fenster: I will see if we can do something. One obvious way is to publish our webpack config to npm and require it from `node_modules`.

Actually, we do **not** need to have any webpack config in the auth library; it's there only for those who want to have a bundle of auth library. Webpack does not read webpack configs of the dependencies, that is, when we webpack the client library, it won't read auth library's webpack config. Hope I made it clear (this comment contains a word 'webpack' 7 times and 'config' 5 times which probably is more than enough).

JustinBeckwith: I think I understand.  So you're saying that the end users of a library need to figure out on their own which dependencies need to be ignored the whole way down through their transitive dependency tree?

alexander-fenster: Yes, that's what I figured. And this is logical. Assume someone wants to use crypto polyfills that are provided by webpack by default; they won't be happy with our decision to strip them.

If a user just needs to use our library (e.g. drive) in the browser app, they use our published webpack config to build a bundle as we suggest. If they want to webpack their own Node.js application that depends on our client library (in terms of `package.json` `dependencies`), it's their job to write the config (they can reuse the one we use, or decide to include some of the modules we removed).

alexander-fenster: (our part of the job here is to make sure the __code__ works - it does; everything related to the bundle size and composition should be taken care of by a person who's making the bundle)

JustinBeckwith: SGTM.

JustinBeckwith:
@alexander-fenster is this PR still needed?

PR #1540, opened 2019-01-10
DO NOT MERGE: example of library usage

alexander-fenster:
**DO NOT MERGE**

This is the suggested format of the micro-packages. Webpackable. 

Two samples are included:

`samples/node/quickstart.js` - quickstart sample (`node samples/node/quickstart.js`)
`samples/web/webserver.js` - webpacked sample (`npm run webpack`, then `node samples/web/webserver.js`).

Please let me know what you think! I'll update the generator then. It can wait till auth library v3.0 release.

### DIFF
--- a/src/apis/drive/package.json
+++ b/src/apis/drive/package.json
@@ -4,16 +4,18 @@
   "description": "drive",
   "main": "build/index.js",
   "types": "build/index.d.ts",
-  "keywords": ["google"],
+  "keywords": [
+    "google"
+  ],
   "author": "Google LLC",
   "license": "Apache-2.0",
   "homepage": "https://github.com/google/google-api-nodejs-client",
   "bugs": {
-    "url" : "https://github.com/google/google-api-nodejs-client/issues"
+    "url": "https://github.com/google/google-api-nodejs-client/issues"
   },
   "repository": {
     "type": "git",
-    "url" : "https://github.com/google/google-api-nodejs-client.git"
+    "url": "https://github.com/google/google-api-nodejs-client.git"
   },
   "engines": {
     "node": ">=6.0.0"
@@ -23,14 +25,19 @@
     "lint": "gts check",
     "compile": "tsc -p .",
     "prepare": "npm run compile",
-    "docs": "typedoc --out docs/"
+    "docs": "typedoc --out docs/",
+		"webpack": "npx webpack"
   },
   "dependencies": {
-    "googleapis-common": "^0.4.0"
+    "googleapis-common": "0.6.0-webpack"
   },
   "devDependencies": {
     "gts": "^0.9.0",
+    "null-loader": "^0.1.1",
+    "ts-loader": "^5.3.3",
+    "typedoc": "^0.14.0",
     "typescript": "~3.2.0",
-    "typedoc": "^0.14.0"
+    "webpack": "^4.28.4",
+    "webpack-cli": "^3.2.1"
   }
 }
--- a/src/apis/drive/samples/.eslintrc.yml
+++ b/src/apis/drive/samples/.eslintrc.yml
@@ -1,0 +1,3 @@
+---
+rules:
+  no-console: off
--- a/src/apis/drive/samples/node/README.md
+++ b/src/apis/drive/samples/node/README.md
@@ -1,0 +1,26 @@
+# Drive v3 API Samples
+
+These samples allow you to list, download, and upload files from Google Drive.
+
+## Running the samples
+
+### **Note: Node.js version 8 or greater is requared to run samples.**
+Set the following values in `oauth2.keys.json` (up one directory):
+
+* `client_id`
+* `project_id`
+* `client_secret`
+
+__Run the `quickstart.js` sample:__
+
+This sample will list the files in the user's Google Drive.
+
+You'll need to do the following before you can run this sample:
+
+* Enable the [Google Drive API](https://console.developers.google.com/apis/api/drive.googleapis.com/overview).
+* Set the redirect API in the Google Cloud Console application registration to `http://localhost:3000/oauth2callback`.
+
+```
+node quickstart.js
+```
+
--- a/src/apis/drive/samples/node/quickstart.js
+++ b/src/apis/drive/samples/node/quickstart.js
@@ -1,0 +1,90 @@
+// Copyright 2018, Google, LLC.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// [START main_body]
+const {drive, auth} = require('@google/drive');
+const express = require('express');
+const opn = require('opn');
+const path = require('path');
+const fs = require('fs');
+
+const keyfile = path.join(__dirname, '../oauth2.keys.json');
+const keys = JSON.parse(fs.readFileSync(keyfile));
+const scopes = ['https://www.googleapis.com/auth/drive.metadata.readonly'];
+
+// Create an oAuth2 client to authorize the API call
+const client = new auth.OAuth2(
+  keys.web.client_id,
+  keys.web.client_secret,
+  keys.web.redirect_uris[0]
+);
+
+// Generate the url that will be used for authorization
+this.authorizeUrl = client.generateAuthUrl({
+  access_type: 'offline',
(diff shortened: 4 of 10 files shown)


src/apis/drive/package.json
```diff
@@ -4,16 +4,18 @@
...
   "types": "build/index.d.ts",
-  "keywords": ["google"],
+  "keywords": [
+    "google"
+  ],
   "author": "Google LLC",
   "license": "Apache-2.0",
   "homepage": "https://github.com/google/google-api-nodejs-client",
   "bugs": {
-    "url" : "https://github.com/google/google-api-nodejs-client/issues"
```
JustinBeckwith: more a nit for the template, but this should be `googleapis` instead of `google` :)

src/apis/drive/samples/.eslintrc.yml
```diff
@@ -0,0 +1,3 @@
+---
```
JustinBeckwith: So this introduces a problem.  Today when we run the generator, we `rm -rf src/apis`, and then generate everything from scratch.  Adding samples into the same directory as the generated code will require a more sophisticated generator that's capable of making updates in place instead of just blasting everything away.  I'm not opposed to that approach, it just needs to be done.

src/apis/drive/samples/node/README.md
```diff
@@ -0,0 +1,26 @@
+# Drive v3 API Samples
```
JustinBeckwith: There's a 0% chance we maintain a readme for this stuff :) If this is auto-generated, fine.  But we can't hand write this stuff.

src/apis/drive/samples/node/quickstart.js
```diff
@@ -0,0 +1,90 @@
+// Copyright 2018, Google, LLC.
```
JustinBeckwith: It's unclear to me if this file would be generated or manual?

src/apis/drive/samples/node/quickstart.js
```diff
@@ -0,0 +1,90 @@
...
+'use strict';
+
+// [START main_body]
+const {drive, auth} = require('@google/drive');
+const express = require('express');
+const opn = require('opn');
+const path = require('path');
+const fs = require('fs');
+
+const keyfile = path.join(__dirname, '..', 'oauth2.keys.json');
```
JustinBeckwith: Thinking out loud - should we rewrite these to use whatever is in `GOOGLE_APPLICATION_CREDENTIALS` for finding the path to the creds file?  That would be consistent with what we do for service account credentials 🤷‍♂️

src/apis/drive/samples/node/quickstart.js
```diff
@@ -0,0 +1,90 @@
...
+this.authorizeUrl = client.generateAuthUrl({
+  access_type: 'offline',
+  scope: scopes,
+});
+
+// Open an http server to accept the oauth callback. In this
+// simple example, the only request to our webserver is to
+// /oauth2callback?code=<code>
+const app = express();
+app.get('/oauth2callback', (req, res) => {
```
JustinBeckwith: I would just go with async code.

src/apis/drive/samples/node/quickstart.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+    client.credentials = tokens;
+    res.send('Authentication successful! Please return to the console.');
+    server.close();
+    listFiles(client);
+  });
+});
+const server = app.listen(3000);
+// open the browser to the authorize url to start the workflow
+opn(this.authorizeUrl, {wait: false});
```
JustinBeckwith: Eugh.  Sadly this sample probably won't exit the process on it's own.  Take a look at the other samples I have that use opn and look at the weird stuff I have to do to shut down express.

src/apis/drive/samples/node/quickstart.js
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Lists the names and IDs of up to 10 files.
+ * @param {auth.OAuth2} auth An authorized OAuth2 client.
+ */
+function listFiles(auth) {
```
JustinBeckwith: asyyyyyyync

src/apis/drive/samples/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "google-drive-samples",
+  "version": "0.0.1",
```
JustinBeckwith: don't need a version for le samples.  Instead should use `private: true`

src/apis/drive/samples/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "These samples allow you to list, download, and upload files from Google Drive.",
+  "main": "quickstart.js",
+  "scripts": {
```
JustinBeckwith: meh, no need to have it

src/apis/drive/samples/web/README.md
```diff
@@ -0,0 +1,36 @@
+# Drive v3 API Samples
+
+These samples allow you to list files on Google Drive using this library in the browser.
```
JustinBeckwith: are we assuming this is hand written or generated?

src/apis/drive/package.json
alexander-fenster: noted!

src/apis/drive/samples/.eslintrc.yml
alexander-fenster: We can put samples elsewhere (or leave them in their current location). I'll see what we can do in the generator though.

src/apis/drive/samples/node/README.md
alexander-fenster: But why? This one was hand-written, I just cropped it. We only have samples for several APIs, we are not going to have more `README` files than number of hand-written samples we support.

src/apis/drive/samples/node/quickstart.js
alexander-fenster: Manual - just like now. Samples will exist only for those APIs they exist for now.

src/apis/drive/samples/node/quickstart.js
alexander-fenster: Looks like a good idea, even though this JSON file is different compared to the service account key.

src/apis/drive/samples/node/quickstart.js
alexander-fenster: Oh yeah. I spent like half an hour trying to make it close and figured I can't, so I just added this `Press Ctrl+C` and relaxed :)

src/apis/drive/samples/node/quickstart.js
alexander-fenster: that's just copied from the existing sample :)

src/apis/drive/samples/package.json
alexander-fenster: it was autogen'd by `npm init`, will remove

src/apis/drive/samples/web/README.md
alexander-fenster: Hand-written

alexander-fenster:
So about handwritten vs generated samples. No way we can generate samples for all APIs, it's something @vchudnov-g is working on for GAPIC libraries and it's unlikely we can do anything about it for this library. I assumed all the samples are hand-written, just like they are now in https://github.com/googleapis/google-api-nodejs-client/tree/master/samples.

But what we can actually do is to have `webserver.js` stuff copied across all APIs to show how to perform OAuth2 workflow with the given API - just have it authenticated and show how to load the given client library in the browser. The whole `webserver.js` and the auth part of `index.html.template` is going to be the same across all APIs. What do you think?

src/apis/drive/samples/.eslintrc.yml
alexander-fenster: OK - after some thinking, I think it will be better to leave samples in the separate folder, at least because of `sampleClient.js` that is shared across all APIs. I suggest that all samples will have `package.json` requiring the corresponding API, just as we do in GAPIC libraries.

src/apis/drive/samples/node/quickstart.js
alexander-fenster: ...but since I'm trying not to change samples too much, I guess this change will be out of scope :)

src/apis/drive/samples/.eslintrc.yml
JustinBeckwith: I actually prefer the other approach 😆 We need to talk this one through in person, because it plays into the whole "how do we auto-generate individual library changes" discussion.  This likely means some stuff with Kokoro.

src/apis/drive/samples/node/README.md
JustinBeckwith: It's going to get out of date.  And without tooling to generate it, there's no way we keep up with them.  I prefer to have no README at all in the samples dir, and put whatever info we need to run the samples up in the top level README.

src/apis/drive/samples/web/README.md
JustinBeckwith: Yeah, I would really rather not :P

JustinBeckwith:
👋 @alexander-fenster is this still a thing?

PR #1547, opened 2019-01-15
feat: generating webpackable packages

alexander-fenster:
Generating small packages that are:
- webpackable
- usable (`AuthPlus` is re-exported from `googleapis-common` to make it available for users).

### DIFF
--- a/src/generator/templates/package.json.njk
+++ b/src/generator/templates/package.json.njk
@@ -4,16 +4,18 @@
   "description": "{{name}}",
   "main": "build/index.js",
   "types": "build/index.d.ts",
-  "keywords": ["google"],
+  "keywords": [
+    "google"
+  ],
   "author": "Google LLC",
   "license": "Apache-2.0",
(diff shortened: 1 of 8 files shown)


alexander-fenster:
@JustinBeckwith This is now ready for review!

src/generator/templates/package.json.njk
```diff
@@ -23,14 +25,19 @@
     "lint": "gts check",
     "compile": "tsc -p .",
     "prepare": "npm run compile",
-    "docs": "typedoc --out docs/"
+    "docs": "typedoc --out docs/",
+    "webpack": "npx webpack"
```
JustinBeckwith: You don't need to use `npx` here since it's installed as a dev dependency and run in the context of an npm script.

alexander-fenster:
This complicated process is exactly the reason why I suggest that generator should live outside of this library, being a devDependency :)

https://github.com/googleapis/google-api-nodejs-client/pull/1554 shows the changes to all APIs.

PR #1573, opened 2019-01-30
fix: respect `.` in api versions (fix adsense API)

### LINKED ISSUE #1531: Invalid parameter v1.4 for adsense
Hey there!

When I try to create an instance of adsense API in a typescript file, only version "v1_4" can be passed. But the correct string is "v1.4" because these is the only one key declared in [VERSIONS variable](https://github.com/googleapis/google-api-nodejs-client/blob/master/src/apis/adsense/index.ts#L20).

So, when I execute `const adsense = google.adsense('v1_4')` the following error is returned:
```
.../node_modules/googleapis-common/build/src/apiIndex.js:36
        throw new Error(`Unable to load endpoint ${api}("${version}"): ${e.message}`);
        ^

Error: Unable to load endpoint adsense("v1_4"): ctr is not a constructor
    at Object.getAPI (.../node_modules/googleapis-common/build/src/apiIndex.js:36:15)
```

#### Environment details

  - Node.js version: 11.0.0
  - npm version: 6.4.1
  - `googleapis` version: 36.0.0

#### Steps to reproduce

  1. Install googleapis
  2. Write in a file.ts `const adsense = google.adsense('v1_4')`
  3. Compile with tsc file.ts
  4. Execute with node the generated file.js

I fixed it, renaming the version "v1_4" by "v1.4" in [this line](https://github.com/googleapis/google-api-nodejs-client/blob/master/src/apis/adsense/index.ts#L23) 

That's all, 

Thanks!

AVaksman:
Fixes #1531

- [x] Tests and linter pass


src/apis/adsense/index.ts
```diff
@@ -20,10 +20,10 @@ export const VERSIONS = {
   'v1.4': adsense_v1_4.Adsense,
 };
 
-export function adsense(version: 'v1_4'): adsense_v1_4.Adsense;
```
JustinBeckwith: This is a generated file!  See `/*! THIS FILE IS AUTO-GENERATED */` :) This looks to be a bug with the generator.

AVaksman: On it! Thanks

test/test.adsense.ts
```diff
@@ -0,0 +1,48 @@
...
+describe('adsense:1.4', () => {
+  let localAdsense: adsense_v1_4.Adsense, remoteAdsense: adsense_v1_4.Adsense;
+
+  before(async () => {
+    remoteAdsense = await Utils.loadApi<adsense_v1_4.Adsense>(
+        googleapis, 'adsense', 'v1.4');
+  });
+
+  beforeEach(() => {
+    localAdsense = googleapis.adsense('v1.4');
```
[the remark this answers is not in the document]
AVaksman: Done
Thanks

AVaksman:
I can't see details on why the test failed.

Issue #1587, opened 2019-02-08
Chat API missing current Event schema


dustinmichels:
Hi! I wish to use the TypeScript types [defined here](https://github.com/googleapis/google-api-nodejs-client/blob/master/src/apis/chat/v1.ts#L592) to write a Hangouts Chat Bot.

It seems the `Event` object is not up to date. It's called `DeprecatedEvent` and differs from the Event my chat bot actually passes. For instance, it says `EventTime` is a string, when I actually receive an object of the form: `{'seconds': number, 'nanos': number`}. (This is also incorrect in the [online documentation](https://developers.google.com/hangouts/chat/reference/rest/v1/Event).)

Are there plans to update this with an interface for `Event` rather than `DeprecatedEvent`?

Thanks!

#### Environment details

  - OS: Windows
  - Node.js version: v10.15.0
  - npm version: 6.4.1
  - `googleapis` version: 37.2.0

#### Steps to reproduce

  1. In a Google Apps Script, [create and deploy a chat bot from template.](https://developers.google.com/hangouts/chat/quickstart/apps-script-bot)
  2. Log the `event` that gets passed to one of the built-in chat bot functions, like `onMessage`
  3. Compare that `Event` object to the one documented online, and [in this node module.](https://github.com/googleapis/google-api-nodejs-client/blob/master/src/apis/chat/v1.ts#L244)

jonathanwiesel:
It seems the issue in the Google tracker is flagged as Wont Fix, ouch

julbrs:
Same issue found here. Seems absurd to mark this issue as wont fix....

Can we ask why it's unfeasible?

![image](https://github.com/googleapis/google-api-nodejs-client/assets/4988590/8a45bd4d-c4d3-4b34-ae37-610a84a96db7)

PR #1588, opened 2019-02-08
change Chat API times from "string" to "TimeObj"

dustinmichels:
Potential fix for https://github.com/googleapis/google-api-nodejs-client/issues/1587

There are several places in the `apis/chat/v1.ts` file where interfaces are declared with some kind of time property with the type "string." When I actually use the API I receive time objects of type `{"seconds": number; "nanos": number}`.

This PR fixes that issues for me, for my own TS usage. Let me know if it looks valid and useful!

- [ ] Tests and linter pass

### DIFF
--- a/src/apis/chat/v1.ts
+++ b/src/apis/chat/v1.ts
@@ -82,6 +82,8 @@ export namespace chat_v1 {
     upload_protocol?: string;
   }
 
+  type TimeObj = { seconds: number; nanos: number };
+
   /**
    * Hangouts Chat API
    *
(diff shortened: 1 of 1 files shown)


dustinmichels:
I signed it!

dustinmichels:
Signed again.

JustinBeckwith:
Greetings, and thank you for the PR!  Sadly - the file you're trying to edit is auto-generated :) All of the stuff we're getting here is based on the backend definition of the service, available here:
https://chat.googleapis.com/$discovery/rest?version=v1

We use that schema definition to generate all of our interfaces and end points (very similar to swagger/openAPI).  It's possible there's a problem on the backend?

dustinmichels:
Gotcha. Is the backend code public? Or is there a different channel for raising this issue?

JustinBeckwith:
The backend is likely private stuff.  The best place to raise this would be from here:
https://developers.google.com/hangouts/chat/

Hope this helps!

PR #1602, opened 2019-02-19
fix: correct types param name for media mime-type

### LINKED ISSUE #1598: Wrong TypeScript type for Drive File Create
#### Environment details

  - OS: OSX Mojave
  - Node.js version: v10.13.0
  - npm version: 6.7.0
  - `googleapis` version: 37.2.0

#### Steps to reproduce

  1. Try to upload a file to drive
  2. Docs say to use `media.mimeType` but the typing says `media.mediaType` is correct
  3. It works when using `media.mimeType` but fails when using `media.mediaType`

Check the docs example (works): https://developers.google.com/drive/api/v3/manage-uploads#convert
See code in question: https://github.com/googleapis/google-api-nodejs-client/blob/79881b09146be564a8d2a0d5fd775ce9adffa19f/src/apis/drive/v3.ts#L2855

❓  Shall I create a PR?

AVaksman:
Fixes #1598

### DIFF
--- a/src/generator/templates/api-endpoint.njk
+++ b/src/generator/templates/api-endpoint.njk
@@ -156,7 +156,7 @@ export interface Schema${{ schema.id }} {
           /**
             * Media mime-type
             */
-          mediaType?: string;
+          mimeType?: string;
 
           /**
             * Media body contents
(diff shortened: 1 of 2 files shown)


AVaksman:
The failing docs test is addressed in PR #1603.

PR #1644, opened 2019-03-19
feat: run the generator

JustinBeckwith:

- [ ] Tests and linter pass

### DIFF
--- a/src/apis/fitness/v1.ts
+++ b/src/apis/fitness/v1.ts
@@ -1740,7 +1740,7 @@ export namespace fitness_v1 {
      * @param {object} params Parameters for request
      * @param {string=} params.endTime An RFC3339 timestamp. Only sessions ending between the start and end times will be included in the response.
      * @param {boolean=} params.includeDeleted If true, deleted sessions will be returned. When set to true, sessions returned in this response will only have an ID and will not have any other fields.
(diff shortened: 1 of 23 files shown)


src/apis/servicenetworking/v1.ts
```diff
@@ -0,0 +1,3036 @@
...
+   * &quot;123456&quot; message { text: &quot;Hi!&quot; })`  The special name
+   * `*` can be used in the body mapping to define that every field not bound by
+   * the path template should be mapped to the request body.  This enables the
+   * following alternative definition of the update method:      service
+   * Messaging {       rpc UpdateMessage(Message) returns (Message) { option
+   * (google.api.http) = {           patch:
+   * &quot;/v1/messages/{message_id}&quot;           body: &quot;*&quot; }; } }
+   * message Message {       string message_id = 1;       string text = 2;     }
+   * The following HTTP JSON to RPC mapping is enabled:  HTTP | gRPC -----|-----
+   * `PATCH /v1/messages/123456 { &quot;text&quot;: &quot;Hi!&quot; }` |
```
bcoe: seems like we're losing formatting on a bunch of comments somewhere along the line; seems like it's probably something that would need to be fixed upstream though?

src/apis/fitness/v1.ts
```diff
@@ -1917,7 +1917,7 @@ export namespace fitness_v1 {
      * millis since epoch). If specified, the API returns sessions modified
      * since this time. The page token is ignored if either start or end time is
      * specified. If none of start time, end time, and the page token is
-     * specified, sessions modified in the last 7 days are returned.
+     * specified, sessions modified in the last 30 days are returned.
```
bcoe: interesting that this changed from 7 to 30 days; is this a breaking change potentially?

JustinBeckwith:
Yeah - pretty much every release in this lib is auto-generated, and most of them are major version.  Will chat more later :)

PR #1649, opened 2019-03-23
Release v38.0.0

ppa7501:

- [ ] Tests and linter pass


JustinBeckwith:
Greetings, and thanks for the PR!  Version 38 of the library is already released.  Is there a specific issue you're trying to see addressed?

PR #1676, opened 2019-04-21
fix(samples): fix readline args in youtube upload

Ajido:
fix to pass correct arguments.


Ajido:
@googlebot I signed it!

PR #1684, opened 2019-04-26
docs(samples): add sample for insertColumn

tejohnso:
Shows how to use the api with [row and column operations](https://developers.google.com/sheets/api/samples/rowcolumn#insert_an_empty_row_or_column).

- [ ] Tests and linter pass


bcoe:
@tejohnso 👋 thank you for the contribution, will work on getting it reviewed soon.

samples/sheets/insertColumn
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const {google} = require('googleapis');
+const sampleClient = require('../sampleclient');
+
```
bcoe: looks like you missed the extension on the filename, let's call this:

`insert-column.js`

If you don't mind?

samples/sheets/insertColumn
```diff
@@ -0,0 +1,47 @@
...
+    }
+  });
+}
+
+const scopes = [
+  'https://www.googleapis.com/auth/spreadsheets'
+];
+
+if (module === require.main) {
+  const [spreadsheetId, sheetId, startIndex, endIndex] = process.argv.slice(2);
```
bcoe: It would be good to tell someone the expected arguments if they do not provide the appropriate # of arguments:

I like the approach taken here:

https://github.com/googleapis/google-api-nodejs-client/blob/master/samples/drive/download.js#L62

bcoe:
@tejohnso does my feedback make sense, would love to get this landed.

bcoe:
@tejohnso I think this is good to go as soon as the linting issue is addressed, here's a patch you could use:

https://github.com/tejohnso/google-api-nodejs-client/commit/9c264b20eb2f49c031b8fcf768c990d4994f8083

tejohnso:
@bcoe Updated.

Is there some way I'm able to cause the checks to run?

PR #1706, opened 2019-05-10
fix: Incorrect case README

### LINKED ISSUE #1701: Incorrect Case readme
Here
https://github.com/googleapis/google-api-nodejs-client#oauth2-client

oauth2Client has different case in different position

IMRahulSaini:
Fixes #1701:  Incorrect Case readme

- [x] Tests and linter pass


bcoe:
@IMRahulSaini thank you for the contribution 👍

PR #1712, opened 2019-05-15
Fix broken options link

Adam13531:
(this is just a README change, so I didn't make an issue or run tests)


bcoe:
@Adam13531 thanks for the contribution :+1: mind signing the CLA if possible, so that we can get this merged?

PR #1731, opened 2019-06-12
docs(samples): add google docs samples

### LINKED ISSUE #1658: Add sample for writing to a Google doc
**Is your feature request related to a problem? Please describe.**
I would like to use a Lambda function to write to an existing Google Doc using the Google API. 

 **Describe the solution you'd like**
An example of how to do this. Lambda-specific would be great, but anything using node would suffice.

 **Describe alternatives you've considered**
Using the Sheets example, but it doesn't fit my use case.

JustinBeckwith:
Fixes #1658.

### DIFF
--- a/samples/docs/create.js
+++ b/samples/docs/create.js
@@ -1,0 +1,70 @@
+// Copyright 2018, Google, LLC.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


JustinBeckwith:
👋 @erickoledadevrel @sqrrrl I am trying to get the most basic google doc creation sample I can rolling.  Having a bit of trouble - the doc gets created, but it gets created with no content (despite my best efforts).  Do you y'all have an example of valid json for the request body of this one around?

sqrrrl:
Unfortunately what you're trying to do isn't currently supported. See https://developers.google.com/docs/api/reference/rest/v1/documents/create

"Creates a blank document using the title given in the request. **Other fields in the request, including any provided content, are ignored.**"

JustinBeckwith:
ahhhhhh - thank you.  I tried *everything* but having good reading comprehension.  Is there a way we could get the `body` element dropped from the doc?  Or have the API throw with a descriptive error if a body is provided?  I can't imagine I'll be the only one to trip on this.

PR #1745, opened 2019-06-24
docs: fixes options section anchor link

abea:
- [✔️] Tests and linter pass
- [✔️] Code coverage does not decrease (if any source code was changed)
- [✔️] Appropriate docs were updated (if necessary)

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ const drive = google.drive({
 });
 ```
 
-See the [Options section][options] for more information.
+See the [Options section][request-options] for more information.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ const drive = google.drive({
 });
 ```
 
-See the [Options section][request-options] for more information.
+See the [Options section](#request-options) for more information.
 
 ## Usage
 


README.md
```diff
@@ -290,7 +290,7 @@ const drive = google.drive({
 });
 ```
 
-See the [Options section][options] for more information.
+See the [Options section][request-options] for more information.
```
bcoe: I think this should be `[Options section](#request-options)`?

abea: I thought so too, but since the original worked like that I assumed it was a valid style. I'll make that update.

JustinBeckwith:
Thanks for the fix!  I looked into this though... and it looks like the link is working today.  I'm going to go ahead and close this out for now.

abea:
It's still not working for me in Firefox or Chrome, taking me to https://github.com/googleapis/google-api-nodejs-client/tree/master#options, which just sits at the top of the page since that's no longer a heading. I'd encourage you to check again, but up to you.

JustinBeckwith:
And I stand corrected ... you were totally right 🙃 Thanks for the fix!

PR #1761, opened 2019-07-11
fix(docs): switch to googleapis.dev link for reference docs

bcoe:


oliverjessner:
It looks like that in the youtube section are the wrong code examples?

https://googleapis.dev/nodejs/googleapis/latest/youtube/index.html

JustinBeckwith:
Greetings @oliverjessner!  Could we trouble you to open a new bug?  If you could call out the specific sample, and link to the page you were using that would be super helpful :)

PR #1798, opened 2019-08-20
fix example drive export

### LINKED ISSUE #1788: example mistake
On [/samples/drive/export.js](https://github.com/googleapis/google-api-nodejs-client/blob/master/samples/drive/export.js) there is a mistake
I tried the function a few times and it downloaded 20kb less each time
So I discover that the function finish to run by await to `res.data.on('end')`
but the stream didn't finish to write the data to the fs yet
The correct function is:
```
async function runSample() {
  // [START main_body]
  const fileId = '1EkgdLY3T-_9hWml0VssdDWQZLEc8qqpMB77Nvsx6khA';
  const destPath = path.join(os.tmpdir(), 'important.pdf');
  const dest = fs.createWriteStream(destPath);
  const res = await drive.files.export(
    {fileId, mimeType: 'application/pdf'},
    {responseType: 'stream'}
  );
  res.data.pipe(dest)
  await new Promise((resolve, reject) => {
    dest
      .on('finish', () => {
        console.log(`Done downloading document: ${destPath}.`);
        resolve();
      })
      .on('error', err => {
        console.error('Error downloading document.');
        reject(err);
      })
  });
  // [END main_body]
}
```

ttv20:
Fixes #1788

- [V] Tests and linter pass
- [V] Code coverage does not decrease (if any source code was changed)
- [V] Appropriate docs were updated (if necessary)

### DIFF
--- a/samples/drive/export.js
+++ b/samples/drive/export.js
@@ -33,18 +33,31 @@ async function runSample() {
     {fileId, mimeType: 'application/pdf'},
     {responseType: 'stream'}
   );
-  await new Promise((resolve, reject) => {
-    res.data
-      .on('end', () => {
-        console.log(`Done downloading document: ${destPath}.`);
-        resolve();
+  await Promise.all([
+      new Promise((resolve, reject) => {
+        res.data
+          .on('end', () => {
+            resolve();
+          })
+          .on('error', err => {
+            console.error('Error downloading document.');
+            reject(err);
+          })
+          .pipe(dest);
+      }),
+      new Promise((resolve, reject) => {
+        dest
+          .on('finish', () => {
+            console.log(`Done downloading document: ${destPath}.`);
+            resolve();
+          })
+          .on('error', err => {
+            console.error('Error saving document.');
+            reject(err);
+          })
       })
-      .on('error', err => {
-        console.error('Error downloading document.');
-        reject(err);
-      })
-      .pipe(dest);
-  });
+    ])
+  }
   // [END main_body]
 }
 


ttv20:
> can't we just switch end to finish I believe? I'd love to keep the change to a minimum and not introduce the Promise.all.
> 
> Thanks for hunting this down.

the 'end' event is the end of the response (`res.data`) and the 'finish' event is the end of the file write stream (`dest`), so it not just switch.
The question is if we address potential error with the response

We can just ignore potential error with the response:
```
async function runSample() {
  // [START main_body]
  const fileId = '1EkgdLY3T-_9hWml0VssdDWQZLEc8qqpMB77Nvsx6khA';
  const destPath = path.join(os.tmpdir(), 'important.pdf');
  const dest = fs.createWriteStream(destPath);
  const res = await drive.files.export(
    {fileId, mimeType: 'application/pdf'},
    {responseType: 'stream'}
  );
  res.data.pipe(dest);
  await new Promise((resolve, reject) => {
    dest
    .on('finish', () => {
      console.log(`Done saving document: ${destPath}.`);
      resolve();
    })
    .on('error', err => {
      console.error('Error saving document.');
      reject(err);
    })
  })
  // [END main_body]
}
```

bcoe:
@ttv20 👋 still want to see this over the finish line? I think you also need to sign the CLA.

ttv20:
> I believe `finish` shouldn't be called on `dest` until the underlying resource has been closed, and has in most cases had all data written to it. tldr; except for really weird edge-cases, I think we can get away with just switching to `finish`?

No. 'finish' is event of `stream.Writable` and 'end' is event of `stream.Readable` so it'll not work to just change to 'finish'
I remember that I tried 'finish' and it was the same and it didn't finish to write the file, I busy this days so I don't have time to test it. (it not spouse to work according the docs, maybe finish is alias of end)

From node doc:
[stream.Readable -> 'end' event:](https://nodejs.org/api/stream.html#stream_event_end)
> The 'end' event is emitted when there is no more data to be consumed **from the stream**.

[stream.Writable -> 'finish' event:](https://nodejs.org/api/stream.html#stream_event_finish)
> The 'finish' event is emitted after the `stream.end()` method has been called, **and all data has been flushed to the underlying system.**

So I think it necessary to wait for `dest` (writable stream) to finish and not miss the writing of the last chunks of data.
And I don't think it's happens only on 'really weird edge-cases', on my case I just run the example on bigger file (700KB file)

bcoe:
@ttv20 thank you for your work on this, it's much appreciated :smile: we want a sample that works!

I just chatted with my colleague @callmehiphop, his recommendation was this simplification:

```js
await new Promise((resolve, reject) => {
  res.data
    .on('error', reject)
    .pipe(dest)
    .on('error', reject)
    .on('finish', resolve);
});
```

I believe this would have the desired effect, where the `on` after the pipe is listening on the `dest` not the `src`.

PR #1824, opened 2019-09-15
fix!: accept null values in request bodies

### LINKED ISSUE #1559: Calendar Typescript Schema$EventDateTime.date type missing null
Using the Google Calendar API, the interface `Schema$EventDateTime` defines `date` without allowing `null`. However, `start.date` must be set to `null` when using `calendar_v3.events.patch()` to convert an all-day event to a time-limited event. The other option is to use the `calendar_v3.events.update()` method but this isn't an option for our implementation.

Definition in code: https://github.com/googleapis/google-api-nodejs-client/blob/c62efb120f9a82084ccb28c6ed56673833cadcb7/src/apis/calendar/v3.ts#L1001

`googleapis` version: googleapis@36.0.0

Code example:
```javascript
const requestBody: calendar_v3.Schema$Event = {
  start: {
    dateTime: newEvent.start_time.toISOString(),
    date: '', // a null is necessary if we need to revert an all-day event back to a time-limited event
  },
  end: {
    dateTime: newEvent.end_time.toISOString(),
    date: '', // a null is necessary if we need to revert an all-day event back to a time-limited event.
  },
};

// "any" hack to get TypeScript to compile
(requestBody as any).start.date = null;
(requestBody as any).end.date = null;

const res = await calendar.events.patch({
    calendarId: googleCalendar.google_id,
    eventId: googleEvent.google_id,
    sendUpdates: 'all',
    requestBody,
}, undefined);
```

EDIT: Added code example

JustinBeckwith:
BREAKING CHANGE: This fixes #1559, and runs the generator.

The following APIs have been removed:
- healthcare/v1alpha2

The following APIs have been added:
- bigqueryconnection/v1beta1
- binaryauthorization/v1
- compute/alpha


acchou:
This change results in many APIs that were typed as `T | undefined` to now have type `T | undefined | null`. For example in `cloudfunctions_v1`:

```
    export interface Schema$ListFunctionsResponse {
        nextPageToken?: string | null;
    }
```

The `nextPageToken?` allows for an implicit `undefined` type, so the complete type is `string | null | undefined`. It's a bit odd to have both `null` and `undefined` as possible results in and off itself, but more troubling is that this is an argument to the `list()` function:

```
    export interface Params$Resource$Projects$Locations$Functions$List extends StandardParameters {
        /**
         * The value returned by the last `ListFunctionsResponse`; indicates that this is a continuation of a prior `ListFunctions` call, and that the system should return the next page of data.
         */
        pageToken?: string;
    }
```

Here, the input type for `pageToken` is `string | undefined`. Is the client supposed to sanitize the `null` value and turn it into `undefined`? Occam's razor would suggest that either `nextPageToken` or `pageToken` has the wrong type here.

There are many type changes, this is just one small example of what might lurk here.

bcoe:
@acchou are you bumping into any specific issues? this is an attempt to address the fact that some APIs do expect a `null` value to be populated, rather than `undefined`.

If you were to populate `null` for the `pageToken`, I would expect the upstream API to either disregards this, or throw an error.

acchou:
> @acchou are you bumping into any specific issues? this is an attempt to address the fact that some APIs do expect a `null` value to be populated, rather than `undefined`.

Then these APIs should have `undefined` removed from their type (i.e. remove the `?`)

> If you were to populate `null` for the `pageToken`, I would expect the upstream API to either disregards this, or throw an error.

The page token is an opaque value, it is given by the return value of one call to the `functions.list` API and passed into a subsequent call. My issue is that the types for the output aren't compatible with the type for the input. Therefore as a user of this API I'm forced to decide how to handle `null`.

It's not a huge deal to work around this specific case, but it is odd that the APIs are defined such that I need to sanitize the value returned by changing `null` to `undefined` for you, even though it's supposed to be an opaque value to me.

The larger issue is that this change creates a fair amount of inconsistency in the API definitions w.r.t. use of `null` vs. `undefined`. This can cause client code to be quite messy, as it needs to navigate this inconsistency in many places.

Issue #1845, opened 2019-10-03
build is currently breaking on master


bcoe:
see: https://github.com/googleapis/google-api-nodejs-client/pull/1840

PR #1846, opened 2019-10-03
fix: remove tests based on urlshortner

bcoe:
`urlshortner`, I just learned your an API, and [now you're gone](https://developers.googleblog.com/2018/03/transitioning-google-url-shortener.html) 😢. Thanks for the years of hard work, providing integration tests for `google-api-nodejs-client`.

Given that we have similar tests based on drive and, I felt we were relatively safe to remove the tests that now fail due to urlshortner lacking a discovery document.

fixes: https://github.com/googleapis/google-api-nodejs-client/issues/1845


bcoe:
@JustinBeckwith I don't love removing tests, but my thinking was that it's good to get the main branch passing; I can take a look tomorrow.

PR #1848, opened 2019-10-04
chore: update tests to rely on blogger rather than urlshortner

### LINKED ISSUE #1845: build is currently breaking on master

bcoe:
updates tests to use blogger API, rather than retired urlshortner API.

fixes #1845


test/test.urlshortener.v1.ts
```diff
@@ -1,102 +0,0 @@
...
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-import * as assert from 'assert';
-import * as nock from 'nock';
-import * as path from 'path';
-import {GoogleApis, urlshortener_v1} from '../src';
-import {Utils} from './utils';
-
-async function testSingleRequest(urlshortener: urlshortener_v1.Urlshortener) {
```
JustinBeckwith: Just to double check - were these tests verifying the specific behaviors of url shortener, or were they covering broad features that aren't tracked elsewhere?  I know it's somewhat unfair to ask that, since I don't even know 👅

bcoe: they seem fairly urlshortner specific, in fact there's a fixture I should also remove.

PR #1889, opened 2019-11-26
test: add test for x-goog-user-project header

bcoe:
Adds a smoke test for `x-goog-user-project` header being populated when calling:

```js
  // This method looks for the GCLOUD_PROJECT and GOOGLE_APPLICATION_CREDENTIALS
  // environment variables.
  const auth = new google.auth.GoogleAuth({
    // Scopes can be specified either as an array or as a single, space-delimited string.
    scopes: ['https://www.googleapis.com/auth/compute']
  });
  const authClient = await auth.getClient();

  // obtain the current project Id
  const project = await auth.getProjectId();

  // Fetch the list of GCE zones within a project.
  const res = await compute.zones.list({ project, auth: authClient });
  console.log(res.data);
```

Depends on https://github.com/googleapis/google-auth-library-nodejs/pull/837 landing, and being pulled into `googleapis-common` as a dependency.


bcoe:
blocked on landing updates to `googleapis-common` and `google-auth-library-nodejs`.

broady:
FWIW if you want to do a smoke test use the translation API and set the quota project to something nonsense and you should get an error back.

PR #1920, opened 2019-12-29
refactor: use explicit mocha imports

JustinBeckwith:


JustinBeckwith:
@bcoe I am absolutely stumped.  I can't imagine how the changes I'm making are causing this test to fail:

```js
1) Transporters
       should add headers to the request from params:
     TypeError: Cannot read property '0' of undefined
      at Interceptor.replyFunction (test/test.transporters.ts:28:56)
      at start (node_modules/nock/lib/playback_interceptor.js:131:26)
```

Nothing I'm touching is even close to impacting that test as far as I can tell, but damn if it doesn't pass on master, and fail in this branch.  Can I phone a friend for an assist here?

JustinBeckwith:
And just like that it works.

PR #1935, opened 2020-01-15
feat!: run the generator

JustinBeckwith:


bcoe:
I don't think we should land https://github.com/googleapis/google-api-nodejs-client/pull/1905, until we figure out why this broke so many tests.

bcoe (changes requested):
seems like this broke a bunch of thing; won't have cycles to dig into it until Friday I don't think.

bcoe (approved):

PR #2020, opened 2020-04-07
test: the kitchen sink system test sometimes times out

bcoe:
`should be able to use the d.ts` occasionally times out, let's start by retrying it.


alexander-fenster:
@JustinBeckwith My experience with similar tests involving `npm install` kind of shows that if it times out, something is going terribly wrong and increasing the timeout won't necessarily help.

bcoe:
@JustinBeckwith we can start with bumping the timeout ... what I'd noticed with some of our other tests that have extremely long timeouts, is that there's a lot of variation in how long they take (and their stability).

PR #2037, opened 2020-04-13
test: use discovery docs from fixture

bcoe:
rather than loading discover docs over the wire, which has been proving flaky (I believe turning net connect on and off is the source of the flakes in #2036), use nock to load discovery docs from a fixture.

fixes: #2036, #2035


bcoe:
bonus, it looks like this should hopefully fix #2035 too.

JustinBeckwith:
nice

PR #2039, opened 2020-04-13
docs: update readme for drive readme

sbatson5:
## What's in this PR?

I was going through the example for google drive and came across the README, where I was told to create `oauth2.keys.json`.  What wasn't immediately clear was that everything needed to be nested under a key of `web`, [as shown here](https://github.com/googleapis/google-api-nodejs-client/blob/master/samples/drive/quickstart.js#L29). A specific example would have been really helpful, so I added a quick one.


sbatson5:
@googlebot I signed it!

bcoe:
@sbatson5 thanks for the contribution 👌 this will be merged as soon as tests pass.

PR #2053, opened 2020-04-22
feat!: run the generator

JustinBeckwith:
@bcoe @alexander-fenster so this was a wild ride.  The root of the test failures had to do with updates in the root url for blogger, which was different in the remote version than the local version.  As a result I had to fix the tests and re-run the generator.  Hey, no breaking changes this time though!


src/apis/bigquery/v2.ts
```diff
@@ -1206,10 +1224,6 @@ export namespace bigquery_v2 {
      * [Optional] The separator for fields in a CSV file. The separator can be any ISO-8859-1 single-byte character. To use a character in the range 128-255, you must encode the character as UTF8. BigQuery converts the string to ISO-8859-1 encoding, and then uses the first byte of the encoded string to split the data in its raw, binary state. BigQuery also supports the escape sequence &quot;\t&quot; to specify a tab separator. The default value is a comma (&#39;,&#39;).
      */
     fieldDelimiter?: string | null;
-    /**
-     * [Optional, Trusted Tester] Deprecated, do not use. Please set hivePartitioningOptions instead.
-     */
-    hivePartitioningMode?: string | null;
```
alexander-fenster, line 1212: Well, formally, this (and the type change below) are breaking. Slightly.

src/apis/cloudidentity/v1beta1.ts
```diff
@@ -119,46 +119,10 @@ export namespace cloudidentity_v1beta1 {
    * Resource representing the Android specific attributes of a Device.
    */
   export interface Schema$AndroidAttributes {
-    /**
-     * Baseband version of Android device.
-     */
-    basebandVersion?: string | null;
```
alexander-fenster, line 125: All these removals too.

alexander-fenster:
I got tired after I read all the way through the letter C, but I would better call it major anyway :)   (also, what's up with the CI?)

JustinBeckwith:
I have a sneaky suspicion that the `rootUrl` for blogger is getting rolled out slowly, and that half the time I get one url and half the time another.  The tests passed locally, then failed locally, then passed here, then failed here.

PR #2056, opened 2020-04-24
feat: generate samples

JustinBeckwith:
This PR removes the code that downloads sample snippets from GCS, and replaces it with an in-house generator.  For reasons I don't fully understand, dartman was only generating samples for a small select number of APIs.  It was easy enough to make this generic, so I rolled it out broadly.  

*The interesting files are at the bottom of the list, so scroll down*

TODO:
- [x] Generate one sample per API method in standalone files
- [x] Get look-behind regex working for projects/my-project paths

After this lands:
- [ ] Update whatever tool cloudsite uses to inline current GCS / disco-artifact-generator samples
- [ ] Delete code in disco-artifact manager that produces nodejs samples

Stuff to think about:
- I dropped the generation of pagination code.  If we care that much, I'll add a pagination feature.
- The plan is to eventually generate standalone samples as well to replace the dartman generated ones used on cloudsite.  All the bits are here, lets just save it for a followup PR.


src/generator/generator.ts
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import * as fs from 'fs';
+import {promises as fs} from 'fs';
```
alexander-fenster, line 14: so... this one is gnarly, it's not allowed with `engines >= 10`, needs 10.something. Maybe stick to `util.promisify` for some more time?

src/generator/generator.ts
```diff
@@ -69,6 +68,33 @@ function isSimpleType(type: string): boolean {
...
+  // remove trailing $
+  if (pattern.endsWith('$')) {
+    pattern = pattern.slice(0, pattern.length - 1);
+  }
+  // replace projects placeholders
+  pattern = pattern.replace('projects/[^/]+', 'projects/my-project');
+  pattern = pattern.replace('locations/[^/]+', 'locations/my-location');
+  // TODO: This should be replaced with a proper regular express
+  // that does a look behind to find the name to be replaced.  Alex
+  // will probably know how to do it.
```
alexander-fenster, line 94: Oh, you don't need look-behind here. You only need look-behinds when the string being looked-behind is unknown. Here, it's easier:

```
> pattern = "^projects/[^/]+/locations/[^/]+/products/[^/]+/referenceImages/[^/]+$"
'^projects/[^/]+/locations/[^/]+/products/[^/]+/referenceImages/[^/]+$'
> pattern.replace(/\^?(\w+)s\/\[\^\/\]\+\$?/g, "$1s/my-$1")
'projects/my-project/locations/my-location/products/my-product/referenceImages/my-referenceImage'
```

You only need to make sure that different regex applications do not overlap, other than that, it's pretty straightforward, with some added backslashes. We just hope they have all plurals ending with "s", I can't singularize in regex properly :)

```
description:   optional ^     project  s   /  [   ^   /   ]  +  optional $
      regex: /     \^?        (\w+)    s  \/ \[  \^  \/  \] \+     \$?      /
```
and then we replace it with `$1s/my-$1` which expands to `projects/my-project`.

src/generator/templates/sample.njk
```diff
@@ -0,0 +1,82 @@
+// Copyright 2020 Google LLC
```
alexander-fenster, line 1: Don't we have a license in a separate included `.njk` as we do in gapic-generator-typescript?
```njk
{% import "_license.njk" as license -%}
{% license.license() -%}
```

src/generator/templates/sample.njk
```diff
@@ -0,0 +1,82 @@
...
+
+  // Acquire an auth client, and bind it to all future calls
+  const authClient = await auth.getClient();
+  google.options('auth', authClient);
+
+  // Do the magic
+  const res = await {{method.id}}({
+    {% if method.parameters %}
+    {% for pname, p in method.parameters|dictsort %}
+    {% if p.description %} // {{ p.description|oneLine|cleanComments|safe }}{% endif %}
```
alexander-fenster, line 51: I don't really like `oneLine`, I see nothing bad if comments spread into multiple lines, like we do in `jsdoc`.

src/generator/templates/sample.njk
```diff
@@ -0,0 +1,82 @@
...
+    
+    {{ pname|cleanPropertyName|safe }}: {% if p.pattern %}'{{ p.pattern|unRegex }}'{% else %}'placeholder-value'{% endif %},
+    {% endfor %}
+    {% endif %}
+    {% if method.request %}
+
+    // Request body metadata
+    requestBody: {
+      // request body parameters
```
alexander-fenster, line 60: Yeah, I guess we need to expand this as well, do we?

src/generator/templates/sample.njk
JustinBeckwith: For you #2057

src/generator/templates/sample.njk
JustinBeckwith: Tend to agree.  Fixed!

src/generator/templates/sample.njk
JustinBeckwith: What do you mean?

src/generator/generator.ts
JustinBeckwith: This worked like a dream.  Thank you!

src/generator/templates/sample.njk
alexander-fenster: I mean, does `requestBody` need to contain some API-specific fields that should be listed in the sample?

JustinBeckwith: We can totally do that.  Turns out it's just *really* annoying :)

src/generator/generator.ts
JustinBeckwith: Only really needed to run the generator, so I'm not too concerned :)

bcoe:
> I dropped the generation of pagination code. If we care that much, I'll add a pagination feature.

I wonder if we could show an async iterator in action, they simplify pagination a ton.

JustinBeckwith:
@alexander-fenster @bcoe this is ready for a look for real

src/apis/abusiveexperiencereport/v1.ts
```diff
@@ -167,6 +167,55 @@ export namespace abusiveexperiencereport_v1 {
     /**
      * abusiveexperiencereport.sites.get
      * @desc Gets a site's Abusive Experience Report summary.
+     * @example
+     * // Before running the sample:
+     * // - Enable the API at:
+     * //   https://console.developers.google.com/apis/api/abusiveexperiencereport
```
alexander-fenster, line 173: This link did not work for me. Not sure if it's something with my account, or the link is wrong. Just FYI.

src/apis/abusiveexperiencereport/v1.ts
```diff
@@ -167,6 +167,55 @@ export namespace abusiveexperiencereport_v1 {
...
+     *   });
+     *
+     *   // Acquire an auth client, and bind it to all future calls
+     *   const authClient = await auth.getClient();
+     *   google.options('auth', authClient);
+     *
+     *   // Do the magic
+     *   const res = await abusiveexperiencereport.sites.get({
+     *     // Required. The name of the site whose summary to get, e.g.
+     *     // `sites/http%3A%2F%2Fwww.google.com%2F`.
```
alexander-fenster, line 195: Is this supposed to be like this, or do we need `| safe`?  (I really don't know)

src/generator/filters.ts
```diff
@@ -0,0 +1,170 @@
...
+
+export function isSimpleType(type: string): boolean {
+  if (type.indexOf('{') > -1) {
+    return false;
+  }
+  return true;
+}
+
+export function cleanPropertyName(prop: string) {
+  const match = prop.match(/[-@.]/g);
```
alexander-fenster, line 40: nit: `/g` is not needed here

src/generator/filters.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @param regex pattern for the given parameter
+ */
+export function unRegex(regex: string): string {
+  // example: ^projects/[^/]+$' ==> projects/my-project
+  let pattern = regex;
+  if (typeof regex !== 'string') {
+    return '';
+  }
+  // remove leading ^
+  if (pattern.startsWith('^')) {
```
alexander-fenster, line 160: The mega-regex in line 168 should take care of leading `^` and trailing `$`.

src/generator/generator.ts
```diff
@@ -431,27 +224,17 @@ export class Generator {
...
+    const apiPath = path.join(srcPath, 'apis', schema.name);
+    const exportFilename = path.join(apiPath, schema.version + '.ts');
     await mkdirp(path.dirname(exportFilename));
+    // populate the `method.fragment` property with samples
+    addFragments(schema);
+    // generate the API (ex: src/apis/youtube/v3.ts)
     await this.render(API_TEMPLATE, {api: schema}, exportFilename);
+    // generate samples on disk at:
+    // src/apis/<service>/samples/<version>/<method>
+    // generateSamples(apiPath, schema);
```
alexander-fenster, line 237: commented out on purpose?

src/apis/abusiveexperiencereport/v1.ts
JustinBeckwith: Ugh.  Updating this to append `.googleapis.com` (not needed in all of them)

src/apis/abusiveexperiencereport/v1.ts
JustinBeckwith: So I *do* use `| safe`.  Turns out this is exactly what's in the source discovery doc 🤦

src/generator/generator.ts
JustinBeckwith: Yes!  The code works, but that's what write the samples to disk.  I don't want to do that during this first pass.

JustinBeckwith:
@bcoe I'm entirely open to adding the pagination stuff with async iterators.  As a first step, we need to hand craft a sample with tests and such.  Any concerns with punting that to a followup PR?

src/apis/abusiveexperiencereport/v1.ts
```diff
@@ -257,6 +306,43 @@ export namespace abusiveexperiencereport_v1 {
...
+     *
+     * async function main() {
+     *   const auth = new google.auth.GoogleAuth({
+     *     // Scopes can be specified either as an array or as a single, space-delimited string.
+     *     scopes: [],
+     *   });
+     *
+     *   // Acquire an auth client, and bind it to all future calls
+     *   const authClient = await auth.getClient();
+     *   google.options('auth', authClient);
```
bcoe, line 329: I like that there are now inline examples, we should double check that our refdoc generator renders these appropriately at deploy time.

src/generator/templates/sample.njk
```diff
@@ -0,0 +1,77 @@
+{% if standalone %}
+{% include "license.njk" %}
+
+// THIS FILE IS AUTO-GENERATED
```
bcoe, line 4: what generates this file?

src/apis/abusiveexperiencereport/v1.ts
JustinBeckwith: Fun fact: inline samples are generally broken in compodoc 😭 
https://github.com/compodoc/compodoc/issues/304

The inline example is no more broken than it was before I guess?  I think @fhinkel is working on a plan to save us

src/generator/templates/sample.njk
JustinBeckwith: Hah, this is hand written - I want the files *it generates* to have that comment.

PR #2061, opened 2020-04-26
build: add an ignore list to the generator

JustinBeckwith:
Sometimes there are APIs that are just broken, like apigee:v1 right now.  Creating a lightweight ignore list so we can opt these out.


src/generator/generator.ts
```diff
@@ -232,14 +232,19 @@ export class Generator {
...
     const res = await this.request<Schemas>({url: discoveryUrl, headers});
     const apis = res.data.items;
     const queue = new Q({concurrency: 10});
     console.log(`Generating ${apis.length} APIs...`);
     queue.addAll(
       apis.map((api: {[key: string]: string}) => {
         return async () => {
-          this.log('Generating API for %s...', api.id);
+          // look at ignore.json to find a list of APIs to ignore
+          if (ignore.indexOf(api.id) >= 0) {
```
alexander-fenster, line 244: we can now use `includes`? :)

src/generator/generator.ts
```diff
@@ -232,14 +232,19 @@ export class Generator {
...
+            return;
```
alexander-fenster, line 245: Log message will be helpful. AFAIR the generator prints APIs it generates, so makes sense to print APIs it ignores as well.

src/generator/generator.ts
JustinBeckwith: You're right!

src/generator/generator.ts
JustinBeckwith: Good call!

PR #2062, opened 2020-04-26
build!: upgrade to gts@2

JustinBeckwith:
BREAKING CHANGE: Node 10 is now the minimum engines requirement.


package.json
```diff
@@ -1,16 +1,18 @@
...
+  "repository": "googleapis/google-api-nodejs-client",
   "license": "Apache-2.0",
-  "dependencies": {
-    "google-auth-library": "^6.0.0",
-    "googleapis-common": "^4.0.0"
+  "description": "Google APIs Client Library for Node.js",
+  "main": "./build/src/index.js",
+  "types": "./build/src/index.d.ts",
+  "engines": {
+    "node": ">=10"
```
alexander-fenster, line 10: this is breaking right?

alexander-fenster: I know I know, we do majors all the time anyway. But still.

samples/gmail/watch.js
```diff
@@ -50,10 +50,7 @@ const scopes = [
 ];
 
 if (module === require.main) {
-  sampleClient
-    .authenticate(scopes)
-    .then(runSample)
-    .catch(console.error);
+  sampleClient.authenticate(scopes).then(runSample).catch(console.error);
```
alexander-fenster, line 53: I liked the old formatting more :)  Maybe `async`/`await` here?

PR #2075, opened 2020-04-30
test: update kitchen test tsc

JustinBeckwith:
I noticed the kitchen tests were busted on an unrelated PR :)

### DIFF
--- a/test/fixtures/kitchen/package.json
+++ b/test/fixtures/kitchen/package.json
@@ -3,13 +3,9 @@
   "version": "1.0.0",
   "description": "An app we're using to test the library. ",
   "scripts": {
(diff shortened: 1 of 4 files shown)


test/fixtures/kitchen/package.json
```diff
@@ -3,22 +3,18 @@
   "version": "1.0.0",
   "description": "An app we're using to test the library. ",
   "scripts": {
-    "check": "gts check",
-    "clean": "gts clean",
```
zamnuts, line 7: this repo still uses transpiled typescript with a `compile` script. i think we should still have these helper scripts: `check`, `clean`, `fix`. if not, what's the equivalency?

PR #2077, opened 2020-05-02
chore: add download command

JustinBeckwith:
There are situations where you want to run the generator, without picking up changes from upstream APIs.  For example - you want to add a new generator feature.  To help pin down the resulting changes, this PR adds a `--use-cache` flag to `generate.js` that will allow it to read the files from a local cache.  There is also a `npm run download` command to trigger downloading and synchronizing just the JSON files associated with discovery. 

*How to review this PR*: I split this into two commits.  One that adds the feature, and one that runs the generator.  The one to review in detail is here:
https://github.com/googleapis/google-api-nodejs-client/pull/2077/commits/d4b37ad18ff8424282347e9eb37ed4d409059048

It's important this PR is *rebase merged* so both commits show up in history.


alexander-fenster:
GitHub is complicated. I added some comments [here](https://github.com/googleapis/google-api-nodejs-client/commit/d4b37ad18ff8424282347e9eb37ed4d409059048) but they are not showing up in the PR :)  Anyway, LGTM with nits.

PR #2079, opened 2020-05-04
build: format code before written

JustinBeckwith:
Running `gts fix` after generation was starting to take several *minutes*.  Instead, this PR runs the code through `prettier` before writing to disk.  Then, `eslint` is run over a single API directory in the tests to keep us honest.


test/test.clients.ts
```diff
@@ -218,6 +219,10 @@ describe('Clients', () => {
     );
   });
 
+  it('should pass eslint for a given client', () => {
+    execSync('npx eslint --no-ignore src/apis/youtube/*.ts');
```
bcoe, line 223: clever, I like this approach of ensuring there aren't regressions.

PR #2096, opened 2020-05-13
feat!: run the generator

JustinBeckwith:


bcoe:
@JustinBeckwith do we want to merge this, or would we rather wait until tomorrow morning and see if the generation works.

JustinBeckwith:
Partner team is asking for it, so lets get this in.  I'm sure there will be changes tomorrow :)

AdriVanHoudt:
Why is this a breaking change? Would be nice if the changelog had more info then 'run the generator' for breaking changes.

PR #2100, opened 2020-05-14
build: ignore revision-only updates

JustinBeckwith:
This change ignores updates to the `discovery` directory if they only contain changes to the `revision` or `etag` fields.  It will reduce the noisiness of re-generation after introducing the discovery caching feature.

One important side effect of this change - as it's written here, *APIs that have been removed from the discovery index will not be automatically cleaned up*.  This is consistent with the approach taken by Go.  I'm on the fence 50/50 if it's a good idea to be honest.  The big benefit I see here is that maybe (just maybe) we can avoid bumping the semver major every time we want to ship.  The downside is that we will never really prune missing APIs.  I can go back and add the code to do the cleanup if folks thing it's advisable.

### DIFF
--- a/src/generator/download.ts
+++ b/src/generator/download.ts
@@ -14,12 +14,10 @@
 
 import * as minimist from 'yargs-parser';
 import * as path from 'path';
-import * as util from 'util';
 import * as fs from 'fs';
 import Q from 'p-queue';
 import {request, Headers} from 'gaxios';
 import {Schemas} from 'googleapis-common';
-import * as rf from 'rimraf';
 import * as mkdirp from 'mkdirp';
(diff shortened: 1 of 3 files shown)


alexander-fenster:
I see no harm in not removing the APIs immediately when they are unpublished. I just want to be able to submit a PR with API removal eventually (I guess there still is a way to regenerate APIs without comparing them with the existing discovery doc, like from scratch, if anyone needs it?)

JustinBeckwith:
Yup!  You can `rm -rf discovery`, re-run the generator, and it will re-download everything from scratch.  That will blank any existing APIs.

PR #2127, opened 2020-05-21
build: add changelog generation

JustinBeckwith:
This continues the slow build towards per-API PRs.  Not hooked up quite yet.

### DIFF
--- a/src/generator/synth.ts
+++ b/src/generator/synth.ts
@@ -61,7 +61,7 @@ export async function synth() {
           prefix = 'feat';
           break;
         case Semverity.MAJOR:
-          prefix = 'feat!:';
+          prefix = 'feat!';
           break;
       }
       console.log(`Submitting change for ${dir}...`);
--- a/src/generator/synth.ts
+++ b/src/generator/synth.ts
@@ -15,15 +15,24 @@ import * as execa from 'execa';
 import * as path from 'path';
 import * as fs from 'fs';
 import * as gaxios from 'gaxios';
+import {Generator} from './generator';
+import {DISCOVERY_URL, ChangeSet} from './download';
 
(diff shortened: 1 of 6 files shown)


src/generator/synth.ts
```diff
@@ -41,13 +50,35 @@ async function main() {
...
+      let prefix: string;
+      switch (semverity) {
+        case Semverity.PATCH:
+          prefix = 'fix';
+          break;
+        case Semverity.MINOR:
+          prefix = 'feat';
+          break;
+        case Semverity.MAJOR:
+          prefix = 'feat!:';
```
bcoe, line 64: I think this won't work for `feat(templates)!: fiddled around with the templates`. Or if someone use the old approach of having `BREAKING CHANGE` in the body.

Throwing two potential ideas out there, I don't love it, but the community parser isn't [that painful to use](https://github.com/googleapis/release-please/blob/master/src/conventional-commits.ts#L175).

Alternatively, I've been threatening to write a minimal parser for about year now, because I hate how overwrought the popular one in the community is ... perhaps we should just bite the bullet and create that dependency?

JustinBeckwith: Good catch!

src/generator/synth.ts
```diff
@@ -15,15 +15,24 @@ import * as execa from 'execa';
...
 
-async function main() {
-  await execa('git', [
-    'config',
-    '--global',
-    'user.email',
-    '"yoshi-automation@google.com"',
-  ]);
-  await execa('git', ['config', '--global', 'user.name', '"Yoshi Automation"']);
+export enum Semverity {
```
alexander-fenster, line 21: That's one good name

PR #2181, opened 2020-05-24
build: enable autodisco for generation

JustinBeckwith:


src/generator/docs.ts
```diff
@@ -66,4 +71,7 @@ async function main() {
   });
   await Promise.all(promises);
 }
-main().catch(console.error);
+
+if (require.main === module) {
```
bcoe, line 75: this feels very Python(y) 😆 

_but, I like the approach for testing._

src/generator/synth.ts
```diff
@@ -30,11 +33,27 @@ export interface Changelog {
   semverity: Semverity;
 }
 
-export async function synth() {
+export const gfs = {
+  rimraf: util.promisify(rimraf),
```
bcoe, line 37: `fs.rmdir` has `recursive: true` now.

bcoe: const {rmdir} = require('fs').promises;
await rmdir('blerg', {recursive: true});

src/generator/synth.ts
```diff
@@ -132,6 +154,17 @@ export function createChangelog(changeSets: ChangeSet[]) {
...
 }
 
+export function getPrefix(semverity: Semverity) {
+  switch (semverity) {
+    case Semverity.PATCH:
+      return 'fix';
+    case Semverity.MINOR:
+      return 'feat';
+    case Semverity.MAJOR:
+      return 'feat!';
```
bcoe, line 164: the spec dictates that `feat(foo)!` would be how you indicate a breaking change, when both types and feature are used.

Why don't you make this a regex rather than a string comparison, then do something like `/feat(\(.*\))?!/`

src/generator/docs.ts
JustinBeckwith: How dare you 😆 There are a few motivations here.  Can you double check my math, and let me know if there's a better way?
- Each of these files are both individually executable, *and* importable as a library
- As you mentioned, testing is nice

Is there a better way to go about it?

src/generator/synth.ts
JustinBeckwith: So this is completely your call, but the docs call it out as experimental / stage 1:
https://nodejs.org/api/fs.html#fs_fs_rmdir_path_options_callback

...  should we do it?

src/generator/synth.ts
JustinBeckwith: Great catch.

src/generator/docs.ts
bcoe: No, I think this is a good approach and will steal it.

src/generator/synth.ts
bcoe: It's been there for 6 months, I feel comfortable using it.

JustinBeckwith: If you don't mind, I'll cut a separate PR where I remove `rimraf` then.  We use it in other places in the codebase, and I'd like to track that all together.

bcoe: yeah, not a blocker.

PR #2188, opened 2020-05-27
fix(types): augment types to manage response streams

### LINKED ISSUE #2052: [TypeScript] The return type of `drive.files.get` is inferred wrongly when downloading as media
The following is a small snippet to download a file from Google Drive using this API.

```typescript
import {createDriveClient} from "./utils";
import fs from "fs";
import * as stream from "stream";
import {promisify} from "util";

(async () => {
    const drive = await createDriveClient();
    const {data} = (await drive.files.get({
        fileId: "anyFileId",
        alt: "media",
    }, {responseType: "stream"})) as unknown as { data: stream.PassThrough };
    const out = fs.createWriteStream("destinationFile");
    await promisify(stream.pipeline)(data, out);
})().catch(console.error);
```

The problem is that `drive.files.get` is defined to be returning (primarily) `GaxiosPromise<Schema$File>` according to the type definitions around [here](https://github.com/googleapis/google-api-nodejs-client/blob/master/src/apis/drive/v3.ts#L3283).  However, when `alt: "media"` is specified, the method returns `string` (by default) or `stream.PassThrough` (when `{responseType: "stream"}`, or maybe other various types.

Therefore, when I try to use the API with TypeScript properly, I have to apply an ugly cast ( `as unknown as { data: stream.PassThrough }` to the response of the function.  Can it be improved?

JustinBeckwith:
Fixes #2052, and adds a lot more TypeScript 😬

### DIFF
--- a/src/generator/templates/api-endpoint.njk
+++ b/src/generator/templates/api-endpoint.njk
@@ -20,10 +20,12 @@ import {
   GoogleConfigurable,
   createAPIRequest,
   MethodOptions,
+  StreamMethodOptions,
   GlobalOptions,
   BodyResponseCallback,
   APIRequestContext,
 } from 'googleapis-common';
+import {Readable} from 'stream';
 
 export namespace {{api.name}}_{{api.version|replace('\.', '_')}} {
 
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   ],
   "dependencies": {
     "google-auth-library": "^6.0.0",
-    "googleapis-common": "^4.1.0"
+    "googleapis-common": "^4.2.0"
   },
   "devDependencies": {
     "@compodoc/compodoc": "^1.1.10",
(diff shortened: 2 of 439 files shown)


src/generator/templates/api-endpoint.njk
```diff
@@ -20,10 +20,12 @@ import {
   GoogleConfigurable,
   createAPIRequest,
   MethodOptions,
+  StreamMethodOptions,
   GlobalOptions,
   BodyResponseCallback,
   APIRequestContext,
 } from 'googleapis-common';
+import {Readable} from 'stream';
```
alexander-fenster, line 28: [suggestion, feel free to ignore]
should this be imported only if used by the given API? (otherwise, unused imports)

I must admit I have no idea if this applies to APIs in this library, but in GAPICs, we do put all such imports inside if statements in templates.

JustinBeckwith: Yeah, in this particular case the import is *only* being used for the `Readable` type.  It could be required for any of the API calls, if the user uses `{responseType: 'stream'}`

PR #2189, opened 2020-05-28
build: add repository dispatch trigger

JustinBeckwith:
Getting the GitHub Action for our cron right has been a bit of a struggle that involves waiting 24 hours and hoping for the best.  This change attempts to fix the current glitch, but adds a repository dispatch hook so a well crafted post with the appropriate auth token can trigger the job.  Firing it looks like this:
https://developer.github.com/v3/repos/#create-a-repository-dispatch-event


bcoe:
TIL: this will allow us to test with a POST?

Issue #2029, opened 2020-04-10
pathParams in Service-client options don't persist


JacobDorman:
#### Environment details

  - Node.js version: `v12.11.1`
  - npm version: `6.14.4`
  - `googleapis` version: `48.0.0`

#### Steps to reproduce
https://github.com/googleapis/google-api-nodejs-client#service-client-options
```js
const gmail = google.gmail({
  version: 'v1',
  auth,

  params: {
    userId: 'me',
  },
})

await gmail.gmail.users.labels.list()
await gmail.gmail.users.labels.list()
```

userId is used in the url `pathParams: ['userId']`, so it is deleted from `Gmail.context` by [googleapis-common](https://github.com/googleapis/nodejs-googleapis-common)

https://github.com/googleapis/nodejs-googleapis-common/blob/0e6d3cfcc89712e2a75ed33ce8ba476b53275016/src/apirequest.ts#L210-L221
```js
  // delete path parameters from the params object so they do not end up in
  // query
  parameters.pathParams.forEach(param => {
    delete params[param];
    if (
      parameters.context &&
      parameters.context._options &&
      parameters.context._options.params
    ) {
      delete parameters.context._options.params[param];
    }
  });
```

related: https://github.com/googleapis/nodejs-googleapis-common/pull/147

bcoe:
@JacobDorman sounds like a bug. Can you workaround by providing the parameter in each list operation?

PR #2190, opened 2020-05-28
test: ensure params are not deleted from global options

JustinBeckwith:
This is expected to fail at this moment because of #2029.  When the change in https://github.com/googleapis/nodejs-googleapis-common/pull/285 lands, and a new version of the module ships, this should start to pass 😎

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   ],
   "dependencies": {
     "google-auth-library": "^6.0.0",
-    "googleapis-common": "^4.2.0"
+    "googleapis-common": "^4.2.1"
   },
   "devDependencies": {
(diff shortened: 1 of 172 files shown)


test/test.options.ts
```diff
@@ -99,17 +99,19 @@ describe('Options', () => {
   });
 
   it('should use the same per-API setting twice', async () => {
```
bcoe, line 101: Can you think of any edge cases we might be missing a test for, e.g., we're happy with our tests for per-request options.

JustinBeckwith: I think we're in pretty ok shape here.  You can always have more tests, but we test a pretty broad range.

PR #2191, opened 2020-05-29
build: set git config before commit

JustinBeckwith:


src/generator/synth.ts
```diff
@@ -60,6 +60,11 @@ export async function synth(options: SynthOptions = {}) {
   if (!token) {
     throw new Error('please include a GITHUB_TOKEN');
   }
+  // only set these while running in the GitHub Actions environment
+  if (process.env.GITHUB_ACTIONS) {
+    await execa('git', ['config', 'user.email', 'yoshi-automation@google.com']);
```
bcoe, line 65: you might consider doing this as a step in the action, this is what I did here:

https://github.com/conventional-changelog/conventional-changelog/blob/master/.github/workflows/ci.yaml#L15

Not a blocker for me, but seems potentially action specific.

PR #2210, opened 2020-06-03
feat: support GoogleAuth parameter

JustinBeckwith:
This change requires https://github.com/googleapis/nodejs-googleapis-common/pull/287 to land before going out.  It does a few things:
- This lets folks pass a `GoogleAuth` parameter.  Anything that has a `request` method with the right signature.  
- Moves `auth` into the `StandardParameters` interface instead of stamping it out every time in each individual generated interface.


src/apis/abusiveexperiencereport/v1.ts
```diff
@@ -40,6 +41,17 @@ export namespace abusiveexperiencereport_v1 {
   }
 
   interface StandardParameters {
+    /**
+     * Auth client or API Key for the request
+     */
+    auth?:
```
alexander-fenster, line 47: nit, but still: the keys were sorted here, maybe `auth` can go below `alt`? :)

JustinBeckwith: All of those standard parameters are actually generated from the discovery file.  This would be silly hard :/

PR #2220, opened 2020-06-05
build: try to use synth for the nightly run again

JustinBeckwith:
The GitHub Action experiment did not work out for a few reasons:

1. It would appear that a PR created by the `github-actions` token will not trigger github actions (CI in this case) to actually run.  This is apparently by design.
2. The next logical step would be to put the `yoshi-automation` personal access token in the secrets for the repository, and use that.  That's a security can of worms I'd prefer to leave unopened.
3. We ran into some issues with OSPO on approving CLA bot to approve PRs from the `github-actions` user account. 

Effectively, relying on GitHub Actions to submit a PR for our purposes is probably a bad idea.  *Instead*, I want to do the same thing from our existing infrastructure, in kokoro.  This will give me the same keys I need.

The only gotcha here is that I am submitting my own PRs from this script, and *do not* want to rely on the autosynth behavior of submitting the change PR for me.  I have a hunch that the branch gymnastics I am doing will lead to no changes getting submitted by autosynth, but we will have to wait and see.


SurferJeffAtGoogle:
A more descriptive comment for the "ugh" commit, please.

JustinBeckwith:
For the `ugh` - good call out :) The intent is for this PR to be squash-and-merged with the original PR title, but it's worth calling out why I did what I did with the second commit.  To successfully run the `npm run submit-prs` command, we need the npm dependencies for the repo installed.

PR #2242, opened 2020-06-17
feat: add disclaimers for services with GAPICs/Veneers

JustinBeckwith:


disclaimers.json
```diff
@@ -0,0 +1,254 @@
...
+  {
+    "api": "documentai",
+    "package": "@google-cloud/documentai"
+  },
+  {
+    "api": "firestore",
+    "package": "@google-cloud/firestore"
+  },
+  {
+    "api": "firestore",
```
alexander-fenster, line 71: is it a typo or do we have two instances of `firestore`?

disclaimers.json
```diff
@@ -0,0 +1,254 @@
...
+  },
+  {
+    "api": "vision",
+    "package": "@google-cloud/vision"
+  },
+  {
+    "api": "webrisk",
+    "package": "@google-cloud/web-risk"
+  }
+]
```
alexander-fenster, line 254: newline :)

src/apis/bigquery/README.md
```diff
@@ -4,6 +4,13 @@
 
 > A data platform for customers to create, manage, share and query data.
 
+*NOTE*: There are multiple npm modules available for this service. This module
+provides a low level HTTP client which exposes the inner workings of the API.
+The [@google-cloud/bigquery](https://www.npmjs.com/package/@google-cloud/bigquery)
+npm module provides a [grpc](https://grpc.io/) based client that is faster, easier to use, and better maintained.
```
alexander-fenster, line 10: `gRPC` would be better

disclaimers.json
JustinBeckwith: So this is interesting!  Technically the content here is just derived from here:
https://github.com/googleapis/google-cloud-node/blob/master/libraries.json

And those are all derived from various repo-metadata files throughout our repositories.  The issues is that we flag multiple libraries (firestore, datastore, storage, logging all come to mind) as using the underlying API.  There's no notion of "the right one".  I have no idea what to do with this problem to be honest.

src/apis/bigquery/README.md
JustinBeckwith: Removed that bit all together :)

Issue #2377, opened 2020-10-01
Following Contributing docs fails locally


elierotenberg:
#### Environment details

  - OS: Windows 10
  - Node.js version: 13.10.1
  - npm version: 6.13.7
  - `googleapis` version: 60.0.1

#### Steps to reproduce

  1. Checkout repository
  2. `npm run samples-test`, `npm run system-test` (without prior `npm run fix`).

Working on a PR, I checkout out the repo and followed the [CONTRIBUTING](https://github.com/googleapis/google-api-nodejs-client/blob/master/CONTRIBUTING.md) instructions.

I ran across several issues, similar to those encountered in [`nodejs-googleapis-common`](https://github.com/googleapis/nodejs-googleapis-common/issues/327).

### Not Windows-related

The following dependencies were already in the [samples `package.json`](https://github.com/googleapis/google-api-nodejs-client/tree/master/samples) but caused `npm run samples test` to fail:
- `@google-cloud/local-auth`,
- `nconf`,
- `express`.

This is because the samples subpackage isn't installed. It is installed as a side effect of `npm run prelint` which may hide the problem, e.g. if `prelint` has run at least once.

Running `npm install` from the `samples` subdir solved it.

### Windows related

The following error is raised when running `npm run samples-test` on Windows:
```
  1) Auth samples
       should support JWT:
     Error: EPERM: operation not permitted, symlink
         'XXXXXX\google-api-nodejs-client\test\fixtures\service.json' ->
         'XXXXXX\google-api-nodejs-client\samples\jwt.keys.json'
```

This is likely because [`fs.symlinkSync`](https://github.com/googleapis/google-api-nodejs-client/blob/master/samples/test/test.samples.auth.js#L44) doesn't work properly on Windows on files (altough it works with directories using `junction` type) or requires privilege elevation.

As a workaround, copying the file instead of symlinking it solved it.

### Authentication related

The problems described in [this issue](https://github.com/googleapis/nodejs-googleapis-common/issues/327) also apply.

bcoe:
@elierotenberg this answers my question 😄 

Is there a tracking issue on Node.js (or is this a known limitation of Windows?). I'm on the Node.js tooling team, and would love  to fix odd things like this.

elierotenberg:
I couldn't find an issue on Node.js, because I believe it "works as intented" on Windows: symlinks in "dir" or "file" mode require elevated privileges, "junction" doesn't but only works on directories.
See also:
[...]

PR #2379, opened 2020-10-01
test: ensure gaxios consumers can use a custom adapter

elierotenberg:

This PR is ahead of #2378.


samples/test/test.samples.auth.js
```diff
@@ -41,7 +41,8 @@ describe('Auth samples', () => {
     const realPath = path.resolve('jwt.keys.json');
     const exists = fs.existsSync(realPath);
     if (!exists) {
-      fs.symlinkSync(fakePath, realPath);
+      const contents = fs.readFileSync(fakePath);
```
bcoe, line 44: curious why we needed to move away from `symlinkSync` to writing a file?

bcoe (changes requested):
Appreciate the additional test 👍 

My only concern with this PR is that I find switching away from a symlink a bit odd, was this failing on your OS?

samples/test/test.samples.auth.js
elierotenberg: See my reply [here](https://github.com/googleapis/google-api-nodejs-client/issues/2377#issuecomment-702694454).
Sorry for spanning the changes in 2 different PRs...

bcoe (approved):

JustinBeckwith (approved):

PR #2410, opened 2020-10-23
docs: experiment with docfx pipeline

JustinBeckwith:
This is the first step towards using docfx for API reference documentation.  There are a fair number of generator changes in here, but mostly cleanup of the output to match what docfx is expecting.  We will still continue using compodoc for the near future for the docs that get published, but we now have a `npm run docs2` command that lets us easily iterate with the docfx config.


package.json
```diff
@@ -19,6 +19,10 @@
     "test": "c8 mocha build/test",
     "predocs": "npm run build-tools",
     "docs": "node build/src/generator/docs",
+    "predocs2": "npm run compile",
+    "docs-extract": "node --max-old-space-size=8192 ./node_modules/@microsoft/api-extractor/bin/api-extractor run --local --verbose",
```
fhinkel, line 23: should just be `api-extractor` instead of the path

package.json
```diff
@@ -19,6 +19,10 @@
...
+    "docs-md": "node --max-old-space-size=8192 ./node_modules/@microsoft/api-documenter/bin/api-documenter markdown --input-folder build/docs --output-folder docs",
+    "docs2": "npm run docs-extract && npm run docs-md",
```
fhinkel, line 25: in other scripts we have this as `docs-devsite`.

fhinkel (approved):
Rubberstamp lgtm. I need to run this to see what warnings we have left

PR #2430, opened 2020-11-06
fix(deps): upgrade to webpack 5 and webpack 4 cli

### LINKED ISSUE #2411: npm install => Cannot find type definition file for 'node'.
Just following the steps from
https://github.com/googleapis/google-api-nodejs-client/tree/master/src/apis/calendar
for building a browser bundle

clone the repo => ok
then
$ cd src/apis/calendar
$ npm install

I get this error :

> @google/calendar@0.1.0 prepare /home/mick/projects/google-api-nodejs-client/src/apis/calendar
> npm run compile

> @google/calendar@0.1.0 compile /home/mick/projects/google-api-nodejs-client/src/apis/calendar
> tsc -p .

> node_modules/gaxios/build/src/common.d.ts:1:23 - error TS2688: Cannot find type definition file for 'node'.

> 1 /// <reference types="node" />

sofisl:
Fixes #2411 
@JustinBeckwith, after doing some digging, I think this is the correct way of handling the upgrade. I think the question that remains is what library we want to use for the polyfills; we could use what webpack was using in v4, which was https://github.com/webpack/node-libs-browser. I am using [pollyfill-library-node](https://www.npmjs.com/package/polyfill-library-node), but that can be easily swapped.


src/generator/templates/webpack.config.js.njk
```diff
@@ -20,18 +20,27 @@ module.exports = {
   entry: './index.ts',
   resolve: {
     extensions: ['.ts', '.js', '.json'],
+    fallback: {
+      crypto: "polyfill-library-node",
+      child_process: "polyfill-library-node",
+      fs: "polyfill-library-node",
+      http2: "polyfill-library-node",
+      buffer: "polyfill-library-node",
```
sofisl, line 28: @sofisl: only want buffer and string for polyfill; rest should be empty

JustinBeckwith:
@sofisl this is looking much better.  A few thoughts:
- I have here is that I have no idea if it actually works.  Without tests it's really hard to tell.  Not your immediate problem, but I wanted to check - did you try this inside one of the API directories, and make sure `npm run webpack` did it's thing?
- Please run `npm run generate -- --use-cache` so we get updated package.json(s) before the next generate command

sofisl:
> @sofisl this is looking much better. A few thoughts:
> 
> * I have here is that I have no idea if it actually works.  Without tests it's really hard to tell.  Not your immediate problem, but I wanted to check - did you try this inside one of the API directories, and make sure `npm run webpack` did it's thing?

I did! And it ran successfully (inside of calendar)
> * Please run `npm run generate -- --use-cache` so we get updated package.json(s) before the next generate command

Sounds good! will do.

sofisl:
@JustinBeckwith, I'm confused about these errors. They seem to be TS errors, I'm not sure what it has to do with the webpack upgrade?

JustinBeckwith (approved):

PR #2458, opened 2020-12-01
feat(actions): switch to GitHub action for releases

bcoe:
This moves us to using a GitHub action for running release-please and publishing to npm, this will make it easier for us to have two release flows in the repo:

1. the top level publication of `googleapis`.
2. the publication of individual libraries inside `src/apis`.

TODO:

- [x] merge https://github.com/googleapis/releasetool/pull/291, so that we stop handling releases with autorelease.
- [x] add kokoro job that publishes docs when we merge  to the main branch.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -24,7 +24,7 @@ jobs:
               await github.issues.addLabels({
                 owner,
                 repo,
-                issue_number: ${{steps.release-please.outputs.pr}},
+                issue_number: ${{steps.release-pr.outputs.pr}},
                 labels: ['autorelease: pending']
               });
               console.log(`Tagged ${{steps.release-pr.outputs.pr}}`)
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -43,9 +43,10 @@ jobs:
         with:
           node-version: 14
           registry-url: 'https://wombat-dressing-room.appspot.com/'
-      - run: |
-          npm install
-          npm publish
+      - name: publish
         if: ${{ steps.tag-release.outputs.release_created }}
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,52 @@
...
+          fork: true
+          package-name: googleapis
+          command: release-pr
+      - id: label
+        uses: actions/github-script@v3    
+        with:
+            github-token: ${{secrets.GITHUB_TOKEN}}
+            script: |
+              const [owner, repo] = process.env.GITHUB_REPOSITORY.split('/');
+              const latestRelease = await github.issues.addLabels({
```
alexander-fenster, line 24: do we need a `latestRelease` variable if it's not used anywhere? maybe just `await github.issues.addLabels(...)` with no assignment?

bcoe, line 24:
```suggestion
              await github.issues.addLabels({
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,52 @@
...
+                owner,
+                repo,
+                issue_number: ${{steps.release-please.outputs.pr}},
```
alexander-fenster, line 27: Is the step called `release-pr`, not `release-please`?

alexander-fenster: (I have no idea actually :) )

bcoe: No, good catch, I believe it should be `release-pr`.

bcoe, line 27:
```suggestion
                issue_number: ${{steps.release-pr.outputs.pr}},
```

alexander-fenster (approved):
Let's see how it works :)

PR #2483, opened 2020-12-08
build: use existing library version

bcoe:
If possible, use the prior version from the package.json. New libraries will default to `0.1.0`.

Addresses https://github.com/googleapis/google-api-nodejs-client/pull/2479, which is a release due to the package.json reverting on the library.

### DIFF
--- a/src/generator/generator.ts
+++ b/src/generator/generator.ts
@@ -230,7 +230,6 @@ export class Generator {
     const packageData = {name: file, desc, version: defaultVersion};
     // Use the version from the existing package.json, if possible:
     try {
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
       const pkgRaw = await readFile(pkgPath, 'utf8');
       const pkg = JSON.parse(pkgRaw);
       packageData.version = pkg.version;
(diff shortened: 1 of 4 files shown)


src/generator/generator.ts
```diff
@@ -211,6 +220,29 @@ export class Generator {
...
+    const pkgPath = path.join(apisPath, file, 'package.json');
+    const packageData = {name: file, desc, version: defaultVersion};
+    // Use the version from the existing package.json, if possible:
+    try {
+      const pkgRaw = await readFile(pkgPath, 'utf8');
+      const pkg = JSON.parse(pkgRaw);
+      packageData.version = pkg.version;
+    } catch (err) {
+      if (err.code === 'ENOENT') {
+        console.info(`${pkgPath} not found`);
```
JustinBeckwith, line 238: Not sure if we need the output here.  This will just happen for any case where there's a new API being shipped.  Should we drop it?

JustinBeckwith (approved):

src/generator/generator.ts
bcoe: @JustinBeckwith I thought it might help us debug if we're finding it's not working, I don't feel strongly if you'd rather cut down on noise.

sofisl (approved):

src/generator/generator.ts
JustinBeckwith: I don't feel strongly either (I put an approve on it)

PR #2517, opened 2021-02-17
chore: update dependencies to support workload identity federation

bojeil-google:
Bump up versions for:
- `google-auth-library` to `^7.0.2`
- `googleapis-common` to `^5.0.1`

The changes above are not sufficient as the `BaseExternalAccountClient` type is still missing in the generated APIs. As a result `BaseExternalAccountClient` was added to the API templates `api-endpoint.njk` and all the APIs were regenerated.
While we're at it, `googleapis-common` dependency in the generated APIs was updated to `^5.0.1`.


bojeil-google:
@bcoe regarding this PR:

- I only updated the templates file and ran `npm run generate`. All the apis file changes were auto-generated. (I didn't manually modify them).
- Regarding the failing npm test, it seems to be due to the test environment running out of memory. I think a rerun could fix it.

bcoe (approved):

PR #2536, opened 2021-03-05
build: add release-please config for submodules

sofisl:


release-please-config.json
```diff
@@ -0,0 +1,137 @@
+{
+  "bootstrap-sha": "6e61af34c0bfdfc3d6f973bffcd6a7e2420590d2",
+  "release-draft": true,
```
bcoe, line 3:
```suggestion
```

bcoe (approved):

.release-please-manifest.json
```diff
@@ -0,0 +1,134 @@
+{
+  "src/apis/abusiveexperiencereport": "0.1.0",
```
JustinBeckwith, line 2: Huh - with `package.json` files available in sub-directories, I wonder what the relationship is between this version and the one there?  Or is this a `semantic-release` type situation where the version in `package.json` is ignored?

.release-please-manifest.json
```diff
@@ -0,0 +1,134 @@
...
+  "src/apis/acceleratedmobilepageurl": "0.1.0",
```
JustinBeckwith, line 3: I don't think we can reasonably keep a file like this up to date unless we have automation to do it.  What creates this file?  What keeps it up to date?

.release-please-manifest.json
```diff
@@ -0,0 +1,134 @@
...
+  "src/apis/vectortile": "0.1.0",
+  "src/apis/verifiedaccess": "0.1.0",
+  "src/apis/webfonts": "0.1.0",
+  "src/apis/webmasters": "0.1.0",
+  "src/apis/workflowexecutions": "0.1.0",
+  "src/apis/youtube": "0.1.0",
+  "src/apis/youtubeAnalytics": "0.1.0",
+  "src/apis/youtubereporting": "0.1.0"
+}
+
```
JustinBeckwith, line 134: Can we get the bonus new-line to make the red circle thingy go away?

release-please-config.json
```diff
@@ -0,0 +1,136 @@
...
+    "src/apis/trafficdirector": {},
+    "src/apis/vault": {},
+    "src/apis/vectortile": {},
+    "src/apis/verifiedaccess": {},
+    "src/apis/webfonts": {},
+    "src/apis/webmasters": {},
+    "src/apis/workflowexecutions": {},
+    "src/apis/youtube": {},
+    "src/apis/youtubeAnalytics": {},
+    "src/apis/youtubereporting": {}
```
JustinBeckwith, line 134: Same questions as above on automation.  Is this generated?  Hand edited?

.release-please-manifest.json
bcoe: The manifest based approach that has been developed treats the manifest as ground truth -- partially because it makes it easier to support many languages, without knowing how to lookup version for every single file for mat (_package.json_, _GemFile_, _setup.py_, etc).

release-please-config.json
bcoe: I assume it will be automated eventually, I'd like to get us to an MVP where we have the existing libraries released.

.release-please-manifest.json
bcoe: I agree that this should be automated, but for the transitional step of getting things working, I'd be comfortable with us landing what we have.

.release-please-manifest.json
bcoe, line 134:
```suggestion
```

bcoe:
@sofisl it might be worth considering building out the automation for generating these manifest files for our next pairing session?

.release-please-manifest.json
JustinBeckwith: Got it.   If that's the case, please change the generator to blank the version number in the deep nested `package.json` files we generate.  In my semantic-release based repos, I set it to `0.0.0` just to signal to folks `aaaaarrrrrgh, here be dragons`.

bcoe:
we have a PR for automation almost read to go.

PR #2545, opened 2021-03-10
test: temporary fix for flaky test related to mocha bug

sofisl:
See: https://github.com/mochajs/mocha/issues/4598
And #2541


JustinBeckwith (approved):
LGTM - lets maybe land this as 'test:' though, since we don't want it in the changelog.

PR #2551, opened 2021-03-16
build: add release please autogeneration for config and manifest for mono-repo publishing

sofisl:


src/generator/generator.ts
```diff
@@ -293,6 +293,59 @@ export class Generator {
     }
     await writeFile(outputPath, output, {encoding: 'utf8'});
   }
+
+  // Generate release-please config for mono-repo publishing
```
bcoe, line 297:
```suggestion
```

I'd like this config to support publishing the root module too, so I'd just drop the comment (the method name is clear).

bcoe (approved):

joeldodge79:
👍

PR #2553, opened 2021-03-18
docs: improve README for refresh tokens

sofisl:


README.md
```diff
@@ -211,6 +212,16 @@ oauth2Client.setCredentials({
 
 Once the client has a refresh token, access tokens will be acquired and refreshed automatically in the next call to the API.
 
+Refresh tokens may stop working after they are granted, either because:
```
JustinBeckwith, line 215: @bshaffer thinking out loud - is there a way a user can hand us a refresh token, and we can tell them if it's valid or not?  Bonus points if we can tell them *why* it isn't valid.  I know [this fella](https://developers.google.com/identity/sign-in/web/backend-auth#calling-the-tokeninfo-endpoint) exists for id tokens, but I have no idea if such a thing could/would work for a refresh token.

README.md
```diff
@@ -211,6 +212,16 @@ oauth2Client.setCredentials({
...
+- The user has revoked your app's access
+- The refresh token has not been used for 6 months
+- The user changed passwords and the refresh token contains Gmail scopes
+- The user account has exceeded a max number of live refresh tokens
+- The refresh token has expired after 7 days
```
bshaffer, line 220: This one doesn't make sense to me. Why would it expire after 7 days? It also seems in conflict with the reason "refresh token hasn't been used in 6 months"

README.md
bshaffer: I do not believe there is any way to tell WHY a refresh token is invalid. There is a oauth "token info" endpoint (https://www.googleapis.com/oauth2/v3/tokeninfo) but it does not accept refresh tokens. 

...and the only way I'm aware of to tell IF it's valid is to try using it to fetch an access token.

README.md
sofisl: You're right, I needed to specify that it's for testing apps! I took these conditions from here: https://developers.google.com/identity/protocols/oauth2

bcoe (approved):

bcoe:
@sofisl @bshaffer @JustinBeckwith given that it sounds like we can't get additional meta information around the refresh_token, I feel like this is sounding pretty good?

JustinBeckwith (approved):

PR #2557, opened 2021-03-18
build: publish packages for each individual API

### LINKED ISSUE #806: Publish packages for each individual API.

bcoe:
**tldr; we now publish individual libraries from the googleapis library**

Here are some examples:

* [@googleapis/youtube](https://www.npmjs.com/package/@googleapis/youtube).
* [@googleapis/sheets](https://www.npmjs.com/package/@googleapis/sheets).

## Why the heck did this take so long?

Gradual work that's been happening in the background to make this possible:

1.  ~2 years ago, the first project I started working on at google was [release-please](https://github.com/googleapis/release-please) (_a tool for automating releases based on commit conventions_). However, my work has concentrated on adding support for 7 languages, vs., the mono-repo work flow required for this repository.
2. ~1 year ago @JustinBeckwith put in work to [add Conventional Commit support to googleapis](https://github.com/googleapis/google-api-nodejs-client/pull/2127). This made it possible for us to consider using release-please for publishing individual modules from googleapis.
3. ~6 months ago rudimentary support for mono-repos was added to [release-please](https://github.com/googleapis/release-please/pull/501) (_the idea being to leverage @JustinBeckwith's work_). But, in experimentation, it didn't work great when applied to a repo with >100 libraries: too many API calls, too many PRs for the maintainer to maintain.
4. ~2 months ago, @joeldodge79 found release-please, and began implementing mono-repo support that allows for releases to be combined into one aggregate PR, vs., many small PRs. Joel built out a feature set [described here](https://github.com/googleapis/release-please/blob/master/docs/manifest-releaser.md) that's perfect for googleapis' needs.
5. ~1 month ago, @sofisl [wrote a tool](https://github.com/sofisl/mono-repo-publish) that allows us to publish these aggregate PRs, using our existing publishing pipeline.
6. This pull request wires everything together.

## What will happen after this lands?

Libraries without a `@google-cloud` version (_[full list here](https://github.com/googleapis/google-cloud-node#google-cloud-nodejs-client-libraries)_) will now be automatically published to npm.

We've already performed an [initial publication for 100s of libraries](https://www.npmjs.com/search?q=%40googleapis).

---

Apologies this took so long, thanks to everyone who contributed to the implementation.

Fixes #806


.github/workflows/release.yaml
```diff
@@ -1,20 +1,18 @@
...
-        uses: GoogleCloudPlatform/release-please-action@v2
+        uses: google-github-actions/release-please-action@v2
         with:
-          token: ${{ secrets.FORKING_TOKEN }}
-          release-type: node
           fork: true
-          package-name: googleapis
-          command: release-pr
+          token: ${{ secrets.FORKING_TOKEN }}
+          command: manifest
```
joeldodge79, line 15: I'm sure you're aware but the way I wrote this it's gonna execute both `manifest-pr` and `manifest-release` (whereas before you just had `release-pr` and not `github-release`)

bcoe, line 15:
```suggestion
          command: manifest-pr
```

bcoe: For this repo, we want `manifest-pr` only (good catch).

.github/workflows/release.yaml
```diff
@@ -1,20 +1,18 @@
 on:
   push:
     branches:
-      - master
+      - remove-release-action-2
```
bcoe, line 4:
```suggestion
      - master
```

sofisl (approved):

manwithsteelnerves:
@bcoe Is it possible for @googleapis/firestore module too (https://github.com/googleapis/google-api-nodejs-client/tree/master/src/apis/firestore) ?

I understand there is a gRpc project of firestore but would like to know if a separate module for Rest api is possible other than doing the below
```
import {
  AuthPlus,
  firestore,
  firestore_v1,
} from 'googleapis/build/src/apis/firestore'
```

JustinBeckwith:
Greetings @manwithsteelnerves!  This is something we're specifically trying to avoid, so we can reduce confusion between the two modules.  That doesn't mean we won't do it, but we'd love to learn more.  Out of curiosity - why do you want REST vs the gRPC based module?

manwithsteelnerves:
@JustinBeckwith Thanks for jumping in!
Its actually for having low startup times. Here is a [long running issue](https://github.com/googleapis/google-cloud-node/issues/2942) regarding cloud functions+firestore.

Also, [here is the on](https://github.com/googleapis/nodejs-firestore/issues/1480#issuecomment-829063598)e recently my tests indicate REST api is much faster compared to gRPC modules. However, as serverless tech is very sensitive to loading times, this became a requirement.

JetUni:
Is there any way the documentation could be updated to reflect this change? For some reason it was quite difficult for me to realize that there were individual modules because all the documents reference the monolithic npm module

PR #2576, opened 2021-03-29
build: add tests for supporting global options

sofisl:


test/test.options.ts
```diff
@@ -127,6 +136,43 @@ describe('Options', () => {
...
+          product,
+          version,
+        },
+      ],
+    });
+    const drive = google.drive({version: 'v2', auth: 'apikey2'});
+    createNockRequestHeaders(
+      '/drive/v2/files/woot?key=apikey3',
+      'user-agent',
+      'product/version google-api-nodejs-client/5.0.2 (gzip)'
```
JustinBeckwith, line 155: Where's the `5.0.2` coming from?

JustinBeckwith (approved):

test/test.options.ts
sofisl: I don't know... this was a part of the headers even without specifying `userAgentDirectives`

bcoe: This value is being populated based on the current version of googleapis-common, and will break when that dependency is updated.

Can we instead switch this to:

```
/product/version google-api-nodejs-client/.* (gzip)/
```

or something like that?

bcoe (changes requested):

bcoe (approved):

PR #2581, opened 2021-04-01
docs: add documentation for submodule publishing & update individual READMEs

### LINKED ISSUE #2567: Document how to install individual APIs
We should document how to install an individual API, such as `@google-apis/youtube`.

Refs: #2557
Refs: #2346 
Refs: #2187

### LINKED ISSUE #2577: Individual API packages description should link to reference documentations
**Is your feature request related to a problem? Please describe.**

I always have trouble finding the API reference specific to the homegraph client, I usually go thru, the following link chain:
https://github.com/googleapis/google-api-nodejs-client → https://googleapis.dev/nodejs/googleapis/latest/ → https://googleapis.dev/nodejs/googleapis/latest/homegraph/index.html (which shows the same README as the main github repo) → https://googleapis.dev/nodejs/googleapis/latest/homegraph/classes/Homegraph.html

 **Describe the solution you'd like**

Now that we have per API packages, ideally the reference would be directly linked from the npm package description: https://www.npmjs.com/package/@googleapis/homegraph

 **Describe alternatives you've considered**

Alternatively per reference could be published to separate set per-API github pages under their own namespace.

sofisl:
Fixes #2577
Fixes #2567


README.md
```diff
@@ -54,6 +54,12 @@ This library is distributed on `npm`. In order to add it as a dependency, run th
 $ npm install googleapis
 ```
 
+Each API is also released as a submodule. In order to add it as a dependency, run the following command, replacing `API_NAME` with the desired api:
```
JustinBeckwith, line 57: I have a few thoughts/questions here:
- We need to pick an option #1, and an option #2.  My initial instinct is to say "you should install the smaller split modules unless you have a reason to choose the big one".  If we agree that's what we want, I'd consider leading with split modules. 
- I would replace `<API_NAME>` with an actual API name.  People are going to blindly copy and paste that thing. 
- Should we consider adding a table here, similar to what we do with google-cloud-node, which lists the available modules?

README.md
```diff
@@ -104,6 +110,25 @@ async function runSample() {
...
 ```
 
+You can also make calls directly to the APIs by installing a submodule:
+
+``` js
+const docs = require('@googleapis/docs')
+
+const client = await docs.docs({
+    version: 'v1',
+    auth: 'YOUR_AUTHENTICATED_CLIENT'
```
JustinBeckwith, line 120: Assume people will just copy and paste these samples.  When you use a string for `auth`, it assumes the service accepts an API key, which most of them don't do.  Instead - can we show something closer to what a user would do for a simple example?

src/generator/templates/README.md.njk
```diff
@@ -18,7 +18,7 @@ To learn more, see [Client Libraries Explained](https://cloud.google.com/apis/do
 ## Installation
 
 ```sh
-$ npm install googleapis
```
JustinBeckwith, line 21: This change is going to generate 180 changes the next time the generator runs.  That may be fine 🤷 Typically when I make a generator change, I like to locally run the generator with `--use-cache`, so all the changes get bundled into one.

README.md
sofisl: > * We need to pick an option #1, and an option #2.  My initial instinct is to say "you should install the smaller split modules unless you have a reason to choose the big one".  If we agree that's what we want, I'd consider leading with split modules.

I think this is fine, but we should probably lead with the root, otherwise we'll need to change a whole lot of documentation pointing to installing `googleapis`, and also because that method is tried and true. Thoughts?

> * Should we consider adding a table here, similar to what we do with google-cloud-node, which lists the available modules?
I was looking into it until I realized that the README is not auto-generated, so we'd need to convert it to a template, add a generation script, etc.... If you think it's worth it, LMK!

src/generator/templates/README.md.njk
sofisl: This method has traditionally made me commit like 500 files... I know I'm doing something wrong, but I'm not quite sure what :)

bcoe (approved):
I'm comfortable letting the nightly generation do its thing, if we go this route. Will leave it to you and @JustinBeckwith to decide.

@sofisl perhaps try running locally with the cache again? the results should only touch the README if you've pulled the latest version.

src/generator/templates/README.md.njk
sofisl: Note to self, don't forget those two middle hyphens:

```
npm run generate -- --use-cache
```

src/apis/abusiveexperiencereport/README.md
```diff
@@ -7,20 +7,20 @@
...
 
 ## Building a browser bundle
 
 This library can be used in a browser. To prepare a single file bundle, clone the
 [repository](https://github.com/googleapis/google-api-nodejs-client) and run
 
 ```sh
-$ cd src/apis/abusiveexperiencereport
-$ npm install
+$ npm install @googleapis/abusiveexperiencereport
```
JustinBeckwith, line 23: ugh, these instructions won't actually work :|    If you are trying to bundle for a browser now ....  I actually don't know what to tell you.  Before this worked because people were responsible for building their own bundles and then copying them over.  Now that we ship a sub module, we either need to:
1. Keep these instructions the same OR
2. Start to run `npm run webpack` as part of our builds, and publish something in the `browser` field in `package.json`


This area isn't my strength uhhh...  @bcoe?

bcoe: I just revisited yargs' webpack config, so was playing a bit with it on the weekend:

https://github.com/yargs/yargs/pull/1900

What if we remove this section entirely for now, but open a tracking issue to add a general `docs/webpack.md` guide at the top level? This could include step by step instructions for webpacking the individual modules (which I think would work similarly to yargs).

JustinBeckwith: +1 to removing the webpack instructions all up.  

For future tracking though - my proposal (I think) is that we do the `npm run webpack` as part of `npm run compile`, and package the results in the npm module.  I'm not 100% sure of if/how `webpack` supports the "browser" field "spec", but if it does we should consider using that instead.

Assume users won't have any idea how to use webpack.

README.md
bcoe: @sofisl @JustinBeckwith there are 100s of APIs is my concern, has the potential to bloat the README.

What if we link to this search query instead: `https://www.npmjs.com/search?q=scope%3Agoogleapis`.

src/apis/abusiveexperiencereport/README.md
bcoe: Opened an issue:

https://github.com/googleapis/google-api-nodejs-client/issues/2584

@sofisl tldr; let's remove the webpack section from the individual READMEs.

sofisl: Done!

README.md
```diff
@@ -54,6 +54,12 @@ This library is distributed on `npm`. In order to add it as a dependency, run th
 $ npm install googleapis
 ```
 
+If you need to reduce startup times, you can also add a submodule as its own dependency. We make an effort to publish submodules that are __not__ in this [list](https://github.com/googleapis/google-cloud-node#google-cloud-nodejs-client-libraries). In order to add it as a dependency, run the following sample command, replacing with your preferred API:
```
bcoe, line 57: nit: install rather than add.

```suggestion
If you need to reduce startup times, you can also install a submodule as its own dependency. We make an effort to publish submodules that are __not__ in this [list](https://github.com/googleapis/google-cloud-node#google-cloud-nodejs-client-libraries). In order to add it as a dependency, run the following sample command, replacing with your preferred API:
```

bcoe (approved):
Looking good to me, left a couple minor nits.

README.md
```diff
@@ -54,6 +54,12 @@ This library is distributed on `npm`. In order to add it as a dependency, run th
 $ npm install googleapis
 ```
 
+If you need to reduce startup times, you can also install a submodule as its own dependency. We make an effort to publish submodules that are __not__ in this [list](https://github.com/googleapis/google-cloud-node#google-cloud-nodejs-client-libraries). In order to add it as a dependency, run the following sample command, replacing with your preferred API:
+
+``` sh
+$ npm install @googleapis/docs
+```
+
```
bcoe, line 62:
```suggestion

You can run [this search](https://www.npmjs.com/search?q=scope%3Agoogleapis) on npm, to find a list of the submodules available.
```

PR #2615, opened 2021-05-21
chore: migrate to owl bot

SurferJeffAtGoogle:


owlbot.py
```diff
@@ -30,6 +23,3 @@
   '.kokoro/publish.sh',
   '.kokoro/release/publish.cfg'
 ])
-subprocess.run(['npm', 'install'])
-subprocess.run(['npm', 'run', 'update-disclaimers'])
```
JustinBeckwith, line 34: We actually need these :) This is what runs the generator nightly, and submits PRs for us.

JustinBeckwith (changes requested):

owlbot.py
SurferJeffAtGoogle: Ok, then Owl Bot cannot do the job.  We need to find another tool.

PR #2651, opened 2021-06-16
build: use GitHub Action for regeneration

JustinBeckwith:
This change moves us from using autosynth to trigger the regeneration job over to a GitHub Action.  The biggest advantage that autosynth gave us was a single GitHub token which had write permissions to the repository.  This made it easier to push changes to a non-main branch, and open PRs with the same token.  With GitHub Actions, we now need two tokens:
- The baked in `GITHUB_TOKEN` can be used to push the changes to an upstream branch
- The `YOSHI_CODE_BOT_TOKEN` has no permissions on this repository, and can be used to open a PR

We have to take this approach because a.) the `GITHUB_TOKEN` cannot open a PR that triggers more actions (like CI), and b.) the `YOSHI_CODE_BOT_TOKEN` by design does not have push permissions to the repository.  

I _think_ this will work, but there's no real way to know until we land this change, and dispatch the action.

### DIFF
--- a/.github/workflows/update-apis.yaml
+++ b/.github/workflows/update-apis.yaml
@@ -1,0 +1,19 @@
+on:
+  schedule:
+    - cron: "0 1 * * *"
+  workflow_dispatch:
+
+jobs:
+  update-apis:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: npm install
+      - run: npm run update-disclaimers
+      - run: npm run submit-prs
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CODE_BOT_TOKEN: ${{ secrets.YOSHI_CODE_BOT_TOKEN }}
--- a/src/generator/synth.ts
+++ b/src/generator/synth.ts
@@ -47,9 +47,13 @@ export async function synth(options: SynthOptions = {}) {
   const statusFiles = status.split('\n').map(x => x.slice(3));
   const apiDir = path.resolve('./src/apis');
   const files = fs.readdirSync(apiDir);
-  const token = process.env.GITHUB_TOKEN;
-  if (!token) {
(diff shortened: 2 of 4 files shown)


.github/workflows/update-apis.yaml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+      - run: npm install
+      - run: npm run update-disclaimers
+      - run: npm run submit-prs
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
SurferJeffAtGoogle, line 18: What populates these secrets?

JustinBeckwith: `GITHUB_TOKEN` is a secret that's automatically available for all GitHub Actions, and has a basic set of permissions on the repository.  More on that here:
https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

`YOSHI_CODE_BOT_TOKEN` - I'll share details offline.

SurferJeffAtGoogle:
> This change moves us from using autosynth to trigger the regeneration job over to a GitHub Action. The biggest advantage that autosynth gave us was a single GitHub token which had write permissions to the repository. This made it easier to push changes to a non-main branch, and open PRs with the same token. With GitHub Actions, we now need two tokens:
> 
> * The baked in `GITHUB_TOKEN` can be used to push the changes to an upstream branch
> * The `YOSHI_CODE_BOT_TOKEN` has no permissions on this repository, and can be used to open a PR
> 
> We have to take this approach because a.) the `GITHUB_TOKEN` cannot open a PR that triggers more actions (like CI), and b.) the `YOSHI_CODE_BOT_TOKEN` by design does not have push permissions to the repository.
> 
> I _think_ this will work, but there's no real way to know until we land this change, and dispatch the action.

Thank you for providing thorough context in the PR description.

SurferJeffAtGoogle (approved):

PR #2701, opened 2021-08-03
chore(nodejs): update client ref docs link in metadata

fhinkel:


bcoe (changes requested):
This link appears to be 404ing, perhaps we don't reference it in the README?

alexander-fenster (approved):

sofisl:
@fhinkel, these links aren't working.

bcoe:
This library has significantly different reference documentation generation, let's revisit this PR later if we decide it's important to have it on `cloud.google.com`.

PR #2829, opened 2021-10-27
fix: ensure standalone generation is npm packable

JustinBeckwith:


feywind (approved):
I'm not super familiar with all the ins and outs of the generators, but looks fine :)

PR #2950, opened 2022-04-14
Update README.md - remove suggestion to use a deprecated package

yuvalm2:
Remove the suggestion of google-cloud since it's deprecated, as may be seen in their site


sofisl:
Hi there! These libraries are not deprecated: https://www.npmjs.com/search?q=%40google-cloud. Feel free to open an issue if you have any questions.

PR #3058, opened 2022-08-31
docs: remove david-dm.org from README

SukkaW:
- [ ] Ensure the tests and linter pass

The `david-dm.org` website is down for 3 months, and the project `alanshaw/david` is not actively maintained anymore. The `david-dm.org` should be considered dead, thus badges should be removed from README.


bcoe (approved):
Thank you for the contribution.

PR #3081, opened 2022-10-03
Fixing code example, which did not work

unlox775:

Fixes #<issue_number_goes_here> 🦕


bcoe:
Hello @unlox775,

These samples are automatically generated, using the library [here](https://github.com/googleapis/discovery-artifact-manager). So, unfortunately if I merge this the change will be immediately overwritten.

Could I instead bother you to open an issue, there must be a bug with the code generator that creates samples.

PR #3205, opened 2023-03-28
chore: record built artifacts 

SurferJeffAtGoogle:


chingor13 (approved):

alexander-fenster (approved):

SurferJeffAtGoogle:
Actually, I think https://github.com/googleapis/google-api-nodejs-client/pull/3206 may be the right way to do this.

SurferJeffAtGoogle:
> Actually, I think #3206 may be the right way to do this.

Nope.  #3206 yielded no changes.

PR #3250, opened 2023-06-06
fix: remove auth test to be compatible with google-auth-library@8.8.0

dansaadati:
There seems to be a broken system test that was working before google-auth-library@8.8.0. This change removes the test as the API no longer seems to be compatible.


danielbankhead:
Are we sure this is an issue with auth 8.8.0 rather than compute? It appears Auth is returning a new `AuthClient` type (`ExternalAccountAuthorizedUserClient`) that could exist, while compute doesn't expect it:

https://github.com/googleapis/google-api-nodejs-client/actions/runs/5163285822/jobs/9301520274?pr=3245#step:7:23

I think the real fix is having compute accept an expanded `AuthClient` type or asserting the type in the test.

PR #3337, opened 2023-09-06
build(package): generate esm output with exports

### LINKED ISSUE #3335: Provide an ESM output target
Currently this project is written in (generates) TypeScript using esm module notation but generates CJS output.

This makes applying size optimizations to the code essentially impossible. While this package is supposed to be used on the server side, there are environments that care about impact of a module on startup time such as serverless environments and CLIs. Measured on my Macbook M1, requiring this packages currently takes about 650ms.

kwonoj:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-api-nodejs-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

Fixes #3335 

This PR attempts to address #3335. Given there are a lot of possible ways to achieve cjs/esm compatible pkg support, I can see this PR is not feasible to this repo's convention and likely can be closed without merge. That's totally fine, however would like to kickstart an initial effort.

Among various approach, this PR nearly mimics recent redux-toolkit's approach (without bundling part) and my personal experiences when I dealt with RxJS (though it's not real esm - https://github.com/ReactiveX/rxjs/blob/master/package.json). Detailed exploration can be found at https://blog.isquaredsoftware.com/2023/08/esm-modernization-lessons/, but in crux this PR does these things

- Preserve `package.json`'s `type`, so existing cjs-based scripts & assumptions (`require..`) works as-is
- Adjust existing cjs output to `build/cjs`
- Introduce another tsconfig to generate esm outputs, generates under `build/esm`. This uses `tsc-multi`(https://github.com/tommy351/tsc-multi), which is a wrapper to generate multiple target outputs with manually transforming extensions per each.
  - Using above tsc-multi, make `build/esm` uses `.mjs` extension to let node.js resolves this as esm explicitly
- Adjust `generators` template to include `.js` extension by default
- Apply conditional exports to `package.json` for `import` / `require` and lastly let default resolves to esm.

Quick local testing confirmed generated pkg can be imported in node.js with cjs / esm (`type:module`), also typescript definitions are being resolved. The caveat for type resolution is it falls back to cjs types always, but it won't be a huge concern since most of types should be equivalent between 2 build outputs. Below's manual testing from https://arethetypeswrong.github.io/ how imports are being resolved.

```
┌───────────────────┬───────────────────────────┬────────────────────────┐
│                   │ "googleapis/package.json" │ "googleapis"           │
├───────────────────┼───────────────────────────┼────────────────────────┤
│ node10            │ 🟢 (JSON)                 │ 🟢                     │
├───────────────────┼───────────────────────────┼────────────────────────┤
│ node16 (from CJS) │ 🟢 (JSON)                 │ 🟢 (CJS)               │
├───────────────────┼───────────────────────────┼────────────────────────┤
│ node16 (from ESM) │ 🟢 (JSON)                 │ 🎭 Masquerading as CJS │
├───────────────────┼───────────────────────────┼────────────────────────┤
│ bundler           │ 🟢 (JSON)                 │ 🟢                     │
└───────────────────┴───────────────────────────┴────────────────────────┘
```

One last thing to note is about conditional exports: conditional exports limits arbitaray imports (including cjs require) to subpath imports, so if anyone used subpath imports like `require('googleapi/somepath/imports')` explicitly it'll likely fail. I'm not certain this should be treated as major breaking - since the entrypoints reexports everything and individual api should be resolvable withput subpath - though.

### DIFF
--- a/generator.md
+++ b/generator.md
@@ -32,7 +32,7 @@ This repository uses [synthtool](https://github.com/googleapis/synthtool/) to re
 You can generate a single API based on a discovery URL. Replace the url and API name below to match the API you'd like to generate:
 ```
 npm run build-tools
-node build/src/generator/generator.js 'https://apigee.googleapis.com/$discovery/rest?version=v1' \
+node build/cjs/src/generator/generator.js 'https://apigee.googleapis.com/$discovery/rest?version=v1' \
   --include-private false \
   --use-cache false
 ```
--- a/package.json
+++ b/package.json
@@ -5,13 +5,33 @@
   "license": "Apache-2.0",
   "description": "Google APIs Client Library for Node.js",
   "main": "./build/cjs/src/index.js",
-  "types": "./build/cjs/src/index.d.ts",
+  "module": "./build/esm/src/index.mjs",
+  "types": "./build/esm/src/index.d.ts",
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "import": {
+        "types": "./build/esm/src/index.d.ts",
+        "default": "./build/esm/src/index.mjs"
+      },
+      "require": {
+        "types": "./build/cjs/src/index.d.ts",
+        "default": "./build/cjs/src/index.js"
+      },
+      "default": {
+        "types": "./build/esm/src/index.d.ts",
+        "default": "./build/esm/src/index.mjs"
+      }
+    }
+  },
   "engines": {
     "node": ">=14.0.0"
   },
   "files": [
     "build/cjs/src",
-    "!build/cjs/src/**/*.map"
+    "!build/cjs/src/**/*.map",
+    "build/esm/src",
+    "!build/esm/src/**/*.map"
   ],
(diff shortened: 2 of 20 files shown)


kwonoj:
I did use these to test resolution locally:

```
//cjs.cjs

const x = require('googleapis');

console.log(require.resolve('googleapis'));
```

```
//esm.mjs
// run with node --experimental-import-meta-resolve esm.mjs

(async () => {
  const googleapi = await import.meta.resolve('googleapis');

  console.log(googleapi)
})();
```

danieljbruce (approved):

danieljbruce (approved):

sofisl:
We discussed this offline, opted instead to add https://github.com/googleapis/google-api-nodejs-client/blob/ea0b1c809c424495391dac8cb78feb6a95f0772d/package.json#L9

PR #3354, opened 2023-09-26
build: add sideEffects: false to package.json

sofisl:
Allows compilers to throw away unused requires.


cramforce (approved):

bcoe (approved):

bcoe (commented):
This LGTM, is there a way that we could write a test for this in the library, so that we confirm it has the desired effect?

mhio:
Should `sideEffects` be boolean `false` ?  esbuild is complaining about it since the 128.0.0 release:

```
▲ [WARNING] The value for "sideEffects" must be a boolean or an array [package.json]

    ./node_modules/googleapis/package.json:9:17:
      9 │   "sideEffects": "false",
        ╵                  ^

1 warning
```

[Webpack documents it](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) similarly as `boolean | string[]`

PR #3411, opened 2023-12-30
feat: include a create permission sample for the drive API

### LINKED ISSUE #3407: Add a permission sample to the Drive V3 samples
Hello!

I have been using the Google API and noticed that there are no samples for changing permissions. I would like to add one!

justinszaro:
- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-api-nodejs-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes #3407


danieljbruce (approved):

danieljbruce (commented):
This seems fine, but someone with more knowledge about this library may know if this is the correct place for these samples to be placed.

PR #3519, opened 2024-07-17
chore: use discovery cache to generate APIs

sofisl:
Resolves b/352084417


danielbankhead (approved):

chingor13 (changes requested):
This updates where the index is downloaded from, but not the individual discovery docs. The urls in the index still point to the real discovery docs.

PR #3520, opened 2024-07-17
chore: update discovery doc generation to discovery-artifact-manager

sofisl:
fixes b/352084417


src/generator/download.ts
```diff
@@ -78,7 +80,7 @@ export async function downloadDiscoveryDocs(
         options.downloadPath,
         api.id.replace(':', '-') + '.json'
       );
-      const url = api.discoveryRestUrl;
+      const url = `${options.discoveryUrl}/${api.id.replace(':', '.')}.json`;
```
chingor13, line 83: I would use api.name + api.version as that is what the discovery cache updater uses to build the filename. I can't guarantee that api.id is the same.

chingor13 (changes requested):
I would separate the change to the download tool from the client update.

chingor13 (changes requested):
The change to the actual discovery document url (`<service>.<version>.json`) appears to have been reverted

chingor13 (approved):

danielbankhead (approved):

PR #3628, opened 2025-02-16
fix: Typo in comment

ihmpavel:
This PR correct minor typo in comment


sofisl:
Unfortunately this is autogenerated so will just be reverted. I'll submit feedback to the API team to change upstream - thanks!

Issue #3418, opened 2024-01-08
gmail.users.threads.get hangs forever


wdanilo:
#### Environment details

  - OS: MacOS Ventura 13.2
  - Node.js version: v21.4.0
  - npm version: 10.2.4
  - `googleapis` version: 105.0.0

#### Steps to reproduce

Reproduction is not simple. Basically:
  1. I have a custom email client which fetches threads from GMail.
  2. I am able to fetch 1867 threads without issues, but there are 3 threads with IDs that I noted down, that hang up `gmail.users.threads.get` forever (it is never resolved nor rejected). 
  3. If I hardcode these 3 IDs in my code to skip them, everything works fine.

I'd be happy to provide additional data if needed. Also, I tested these IDs on the Google API website API tester and it provides correct result, so it seems like this is an issue with this library.

wdanilo:
Actually, I just tested my program without 

```
google.options({
      http2: true,
    })
```

An it appears that it works. It seems that `gmail.users.threads.get` hangs on these threads only if `http2` is in use.

wdanilo:
Possibly related: https://github.com/googleapis/google-api-nodejs-client/issues/3108

wdanilo:
Any progress here? :)

gring2:
I got same issue, when i fetch calendar events in tight loop with http2 options. It looks like there are problem in http2 stream.

PR #3894, opened 2026-02-12
fix: fall back to HTTP/1.1 when global http2 option is enabled

Deeven-Seru:
## Summary
This change mitigates hanging requests reported in #3418 when `google.options({http2: true})` is used.

## Root cause
The hang occurs in the HTTP/2 request path (`googleapis-common`) where some GET requests can remain open indefinitely. Since this repo surfaces that path through global options, affected calls appear to hang forever.

## What changed
- Added normalization of global options in `GoogleApis.options()`.
- If `http2: true` is provided globally, emit a warning and force `http2: false` (HTTP/1.1 fallback).
- Preserved all other global options unchanged.
- Added tests to verify fallback behavior and warning emission.

## Tests
- `npx mocha build/test/test.options.js --grep "disable http2|not emit warning"`
- `npx mocha build/test/test.options.js`

## Notes
This is a defensive mitigation in `google-api-nodejs-client` until a full HTTP/2 transport fix is available downstream.

### DIFF
--- a/src/googleapis.ts
+++ b/src/googleapis.ts
@@ -62,7 +62,29 @@ export class GoogleApis extends GeneratedAPIs {
    * @param options - Configuration options.
    */
   options(options?: GlobalOptions) {
-    this._options = options || {};
+    this._options = this.normalizeGlobalOptions(options);
+  }
+
+  private normalizeGlobalOptions(options?: GlobalOptions): GlobalOptions {
+    if (!options) {
+      return {};
+    }
+
+    if (!options.http2) {
+      return options;
+    }
+
+    process.emitWarning(
+      'google.options({http2: true}) is temporarily disabled due hanging request issues. Falling back to HTTP/1.1.',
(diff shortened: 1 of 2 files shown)


quirogas:
HI @DEVELOPER-DEEVEN , thank you for taking a swing at this bug. unfortunately, the code you are modifying is autogenerated, so the change has to happen upstream. Additionally, falling back to http/1.1  comes with it's own set of security concerns, so we would probably want to implement a more secure fall back.

Deeven-Seru:
Thanks for the clarification @quirogas . I missed that these files were autogenerated—makes total sense why this can't be merged here.
Regarding the HTTP/1.1 fallback, I understand the security/stability implications of a global downgrade. I'll look into googleapis-common or the upstream generator to see if there's a safer way to handle the hanging request timeouts without forcing a protocol downgrade.
 Thanks for the review!

PR #3901, opened 2026-03-02
feat: run generator manually

danieljbruce:
## Description

The typescript generator PR at https://github.com/googleapis/google-api-nodejs-client/pull/3890 is blocked by some compiler errors so this PR adds a few extra changes that allow it to pass CI checks. In particular, it looks like the lifesciences package was merged even though it was an incomplete package. Another casting error was addressed.

## Impact

Allows us to add generator changes without breaking the library


shivanee-p (approved):

sofisl:
Wondering what this PR actually does - is it just manually running the download step and then making the necessary changes? I think this could work to unblock, but the real issue still stands. The issue is that lifesciences was completely deprecated, so the generator needs to make sure it completely deletes the rest of lifesciences from being exported in the package.

sofisl (approved):

PR #3946, opened 2026-07-28
chore: update test matrix to cover currently supported node versions

pearigee:
Unit tests are currently breaking with the following error:

```
Run npm install --production --engine-strict --ignore-scripts --no-package-lock
npm warn config production Use `--omit=dev` instead.
npm error code EBADENGINE
npm error engine Unsupported engine
npm error engine Not compatible with your version of node/npm: brace-expansion@5.0.8
npm error notsup Not compatible with your version of node/npm: brace-expansion@5.0.8
npm error notsup Required: {"node":"20 || >=22"}
npm error notsup Actual:   {"npm":"10.8.2","node":"v18.20.8"}
npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2026-07-28T20_43_01_654Z-debug-0.log
```

We could implement an override, but I think it makes sense to simply drop the older/unsupported runtimes as a test target. This will also match the test targets in our primary repo, google-cloud-node (as of https://github.com/googleapis/google-cloud-node/pull/8993/changes).


pearigee:
It looks like ci.yaml has already drifted significantly from the template in synthtool. It's not clear to me yet when/if this file would ever be regenerated.

https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/node_library/.github/workflows/ci.yaml

PR #3979, opened 2026-08-26
fix(generator): delete API directories that have no version files left

danieljbruce:
Please ignore. This is just a test to see what the generator does.


danieljbruce:
image

This change doesn't actually compile when added to yoshi bot's "run the generator" branch.

PR #3986, opened 2026-08-28
test: Add bun testing for apiary

danieljbruce:
## Description

This pull request does for apiary what https://github.com/googleapis/google-cloud-node/pull/9208/changes did for google-cloud-node. It adds a CI test that uses `bun run test` instead of `npm test`

Adds a new GitHub Actions presubmit workflow (`.github/workflows/presubmit-bun.yml`) to run the unit test suite against the [Bun](https://bun.sh/) JavaScript runtime across Node versions 22, 24, and 26.

Key details:
- Sets up Bun `1.3.14` via `oven-sh/setup-bun` pinned to SHA `0c5077e51419868618aeaa5fe8019c62421857d6` (v2.2.0).
- Uses `npm install` for dependency management to ensure lockfile and version consistency between CI environments.
- Executes tests with `bun run test`.

## Impact

Ensures ongoing compatibility with the Bun runtime and catches runtime-specific regressions early for users running `googleapis` in Bun environments. This change is CI-only and introduces no breaking changes or runtime library modifications.

## Testing

- Added `.github/workflows/presubmit-bun.yml` to run on pull requests.
- Validated test matrix against Node 22, 24, and 26 with Bun 1.3.14.

## Additional Information

- Dependencies are installed with `npm` rather than `bun install` to preserve package resolution consistency with the existing CI pipeline.
- Sets `MOCHA_THROW_DEPRECATION: false` consistent with existing test jobs.


feywind (approved):

feywind:
I wonder if we should change the branch out of date check here like we did on google-cloud-node.

danieljbruce:
> I wonder if we should change the branch out of date check here like we did on google-cloud-node.

That seems like a good idea. I don't think there are any risks?

PR #3994, opened 2026-09-02
fix: apiary upgrade dependencies to node 22

danieljbruce:
## Description

This pull request completes the minimum node version upgrade to 22 along with making sure all the upgrades dependencies match the node 22 version. It checks in a script that was used to make the upgrade and then removes the script from source control so that we can reference it later. This script was used to upgrade all the dependencies in package.json files. We also migrated the testing infrastructure from Node 18 to Node 22.

## Impact

Ensures apiary dependencies will get all the latest patches going forward.


danieljbruce:
Paying attention to that warning and removing the Node 18 folder will be done in https://b.corp.google.com/issues/556218269. Next we will:
1. Add a changelist to point kokoro runs to the new config files
2. Get access on demand to run the right kokoro jobs
3. Write a PR to eliminate the node 18 folder

shivanee-p (approved):

PR #4004, opened 2026-09-04
fix: rollback generator

danieljbruce:
## Description

https://github.com/googleapis/google-api-nodejs-client/pull/3992 is being rolled back because these changes must be re-merged with a feat! commit message so that all libraries get a major version upgrade in the release PR so that users upgrading to Node v22 must intentionally opt-in to avoid breaking changes.

## Impact

Fix the release to rollout major versions for Node 22.

## Additional notes

We are going to revert these changes in a subsequent PR.


bshaffer (changes requested):
If you put this in the PR description of the already merged PR and then you regenerate the release please PR, it should update the conventional commit:

```
BEGIN_COMMIT_OVERRIDE
feat!: foo bar baz
END_COMMIT_OVERRIDE
```

We should try that first before rolling things back, if the only goal is to change the conventional commit

danieljbruce:
> If you put this in the PR description of the already merged PR and then you regenerate the release please PR, it should update the conventional commit:
> 
> ```
> BEGIN_COMMIT_OVERRIDE
> feat!: foo bar baz
> END_COMMIT_OVERRIDE
> ```
> 
> We should try that first before rolling things back, if the only goal is to change the conventional commit

I found this doesn't work because this was actually merged in a commit sent out in the last release. This pull request rolls back the breaking changes (the node engine upgrade) so we should merge these changes and then cut another minor release.

bshaffer (approved):
Yes, I didn't understand that the releases had already gone out. I agree that reverting them, cutting a new release with the reverted changes, and then marking the previous release as `deprecated` is the best policy here

